PR #2, opened 2018-01-10
Add RethinkDB-based storage for FSM

Ars2014:


aiogram/contrib/fsm_storage/redis.py
```diff
@@ -377,8 +377,10 @@ def has_bucket(self):
     :param storage2: instance of RedisStorage2
     :return:
     """
-    assert isinstance(storage1, RedisStorage)
-    assert isinstance(storage2, RedisStorage2)
+    if not isinstance(storage1, RedisStorage):  # better than assertion
```
Ars2014: After compilation 'assert expressions' being skipped.

PR #11, opened 2018-02-03
Added left_chat_member state to the Message class

apik:
Supplementary state `left_chat_member` for existing state `new_chat_members`.
Maybe useful to monitor if the bot was kicked from the group chat.

### DIFF
--- a/aiogram/types/message.py
+++ b/aiogram/types/message.py
@@ -101,7 +101,7 @@ class Message(base.TelegramObject):
         if self.new_chat_members:
             return ContentType.NEW_CHAT_MEMBERS[0]
         if self.left_chat_member:
-            return ContentType.LEFT_CHAT_MEMBER
+            return ContentType.LEFT_CHAT_MEMBER[0]
         if self.invoice:
             return ContentType.INVOICE[0]
         if self.successful_payment:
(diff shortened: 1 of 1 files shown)


aiogram/types/message.py
```diff
@@ -673,6 +676,7 @@ class ContentType(helper.Helper):
     LOCATION = helper.ListItem()  # location
     VENUE = helper.ListItem()  # venue
     NEW_CHAT_MEMBERS = helper.ListItem()  # new_chat_member
+    LEFT_CHAT_MEMBER = helper.Item()  # left_chat_member
```
JrooTJunior: In this place need to use `helper.ListItem()` instead of `helper.Item()`
`Item` returns string
`ListItem` returns list

aiogram/types/message.py
```diff
@@ -100,6 +100,8 @@ def content_type(self):
             return ContentType.VENUE[0]
         if self.new_chat_members:
             return ContentType.NEW_CHAT_MEMBERS[0]
+        if self.left_chat_member:
+            return ContentType.LEFT_CHAT_MEMBER
```
JrooTJunior: Add item getter: `[0]`

PR #15, opened 2018-02-17
Using connection pool in RethinkDB driver

Ars2014:


JrooTJunior:
Can you add configurable connections limit?

PR #21, opened 2018-03-15
Minor fix

surik00:
Fix None check for response from handlers
+ Minor typos fix


JrooTJunior:
Also i think you can add `return self` in `insert` method too.
And similarly make it for an `ReplyKeyboardMarkup`.

surik00:
Done

Now you can do so:
```py
kb = ReplyKeyboardMarkup().add(KeyboardButton('TEXT1')).row(
    KeyboardButton('1'),
    KeyboardButton('2'),
).insert(KeyboardButton('TWOO'))
```

PR #22, opened 2018-03-18
Split WEBHOOK and WEBAPP settings

Olegt0rr:


MkadirDev:
The webapp is not working as expected for android devices, it works fine on laptop and ios devices. Has anyone faced such a problem?

Olegt0rr:
@MkadirDev , PPR is not a good place to answer questions. Use [discussions](https://github.com/aiogram/aiogram/discussions). And add more info about your issues

Issue #26, opened 2018-03-24
Handler doesn't handle Venue Messages


soermejo:
```
@dp.message_handler(content_types=types.ContentType.VENUE)
async def func(message):
    print(message)
```
i tried this code, and it doesent handler Venue messages

if i use a filter on any types like
`@dp.message_handler(content_types=types.ContentType.ANY)` 
This handle Venue in addition to others

JrooTJunior:
Fixed in latest version: https://github.com/aiogram/aiogram/releases/tag/v1.2.1
You can install new version from pypi: `pip install aiogram -U`

PR #27, opened 2018-03-25
Swap around checking of VENUE and LOCATION content

Olegt0rr:
Fix for issue https://github.com/aiogram/aiogram/issues/26

Venue Message consists of location and venue fields. 
If you check location first - you always set message type = location

### DIFF
--- a/aiogram/types/message.py
+++ b/aiogram/types/message.py
@@ -95,10 +95,10 @@ class Message(base.TelegramObject):
             return ContentType.VOICE[0]
         if self.contact:
             return ContentType.CONTACT[0]
-        if self.location:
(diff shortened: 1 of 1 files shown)


aiogram/types/message.py
```diff
@@ -95,10 +95,10 @@ def content_type(self):
             return ContentType.VOICE[0]
         if self.contact:
             return ContentType.CONTACT[0]
-        if self.location:
-            return ContentType.LOCATION[0]
         if self.venue:
```
Senpos: Since **Venue** contains `location` **and** `venue` properties, isn't it better to check for both of them, rather then rely on their positioning in the code?
`if self.venue and self.location`

PR #48, opened 2018-06-10
Babel is not in requirements

surik00:
No need to check for babel because it is in requirements now


surik00:
Fixed the docstring

PR #55, opened 2018-07-17
Simple example with webhook executor

Olegt0rr:
Simple example with webhook executor


JrooTJunior:
Accepted but in future will be changed

PR #56, opened 2018-07-21
Update json.py to fix pyCharm warning

Olegt0rr:
fix pyCharm warning
'ujson' in try block with 'except ImportError' should also be defined in except block
This inspection detects names that should resolve but don't. Due to dynamic dispatch and duck typing, this is possible in a limited but useful number of cases. Top-level and class-level items are supported better than instance items.

### DIFF
--- a/aiogram/utils/json.py
+++ b/aiogram/utils/json.py
@@ -3,11 +3,10 @@ import json
 try:
     import ujson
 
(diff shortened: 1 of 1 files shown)


JrooTJunior:
Already changed in branch`dev-2.x`: afbe7bb

PR #80, opened 2018-11-15
Check error handler callback args to prevent cycled exceptions

Olegt0rr:
Fixes issue with CPU overloading and memory leaking with more than 2 args in errors handler callback.
Added error handler callback args amount checking to prevent cycled exceptions.


JrooTJunior:
This PR is still actual?

PR #88, opened 2018-12-04
Faster skip updates (calling getUpdates with a negative offset)

Olegt0rr:
Bot can "clean" the updates by calling getUpdates method with a negative offset


JrooTJunior:
Result of Dispatcher.skip_updates is used in executor log message. Need to fix this.
https://github.com/aiogram/aiogram/blob/dev-2.x/aiogram/utils/executor.py#L341-L343

Olegt0rr:
> Result of Dispatcher.skip_updates is used in executor log message. Need to fix this.
> https://github.com/aiogram/aiogram/blob/dev-2.x/aiogram/utils/executor.py#L341-L343

Done!

PR #102, opened 2019-02-22
Fixed entities parsing for captions

Kylmakalle:
# Description

Added specific checks for captions_entities to parse

## Type of change

# How Has This Been Tested?

Manually


aiogram/types/message.py
```diff
@@ -193,7 +193,7 @@ def parse_entities(self, as_html=True):
 
         quote_fn = md.quote_html if as_html else md.escape_md
 
-        if not self.entities:
+        if not (self.entities or self.caption_entities):
```
JrooTJunior: You can move that to variable (`entities = ...`) and then use it in other places

PR #105, opened 2019-02-25
Check bot instance on Dispatcher __init__

Olegt0rr:
If user provides wrong data to 'bot' argument, raised exceptions are not clear to fix issue.

### DIFF
--- a/aiogram/dispatcher/dispatcher.py
+++ b/aiogram/dispatcher/dispatcher.py
@@ -35,6 +35,9 @@ class Dispatcher(DataMixin, ContextInstanceMixin):
                  throttling_rate_limit=DEFAULT_RATE_LIMIT, no_throttle_error=False,
                  filters_factory=None):
 
+        if not isinstance(bot, Bot):
+            raise TypeError(f"Argument 'bot' must be an instance of Bot, not '{type(bot).__name__}'")
+
         if loop is None:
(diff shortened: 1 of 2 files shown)


aiogram/dispatcher/dispatcher.py
```diff
@@ -35,6 +35,9 @@ def __init__(self, bot, loop=None, storage: typing.Optional[BaseStorage] = None,
                  throttling_rate_limit=DEFAULT_RATE_LIMIT, no_throttle_error=False,
                  filters_factory=None):
 
+        if not isinstance(bot, Bot):
+            raise TypeError(f"Argument 'bot' must be an instance of Bot, not '{type(bot).__name__}'")
+
         if loop is None:
             loop = bot.loop
```
Olegt0rr: This call of bot.loop can raise exception if bot argument is not Bot instance.

aiogram/dispatcher/dispatcher.py
```diff
@@ -276,7 +279,7 @@ def stop_polling(self):
 
         :return:
         """
-        if self._polling:
+        if hasattr(self, '_polling') and self._polling:
```
Olegt0rr: If exception occurs before `self._polling` assign - here will be raised another one exception.
That's why I add `hasattr()` check.

PR #110, opened 2019-03-20
Added timeout for every method.

### LINKED ISSUE #109: Adding timeout for Bot API methods.
**Is your feature request related to a problem? Please describe.**
The problem is that Bot API doesn't send any errors, if bot got limited with some methods, like delete_message. Example code:
```
l=[(i, -1000-i) for i in range(1000)]
for i in l:
   try:
      await bot.delete_message(message_id=max(i), chat_id=min(i))
   except Exception:
       pass
```

If there are no `await asyncio.sleep()`, a bot will freeze after deleting ~700-800 msgs. No exceptions will raise, but the process will be completely blocked.

**Describe the solution you'd like**
The best way to solve this problem is by adding timeout for every method, e.g. in `aiogram.Bot` params. 

**Additional context**
[Example](https://t.me/aiogram_ru/44012): `INFO:bot:Process update [ID:711526853]: [success] (in 15379 ms)`

Forden:
# Description

This commit adds timeouts for every API method, that saves your bot from being stuck because of Telegram API unexpected behavior.

Fixes #109  (issue)

## Type of change


gabbhack:
Using the default None we leave this task to the user. I think we should come up with an adequate timeout by default.

JrooTJunior:
Why timeout argument was added only for methods in Bot class? What about shorted API methods (`message.reply` and etc.)?

JrooTJunior:
I think about setting timeout by the context manager instead of passing as argument it to the all of API methods.

Example:
``` python3
with bot.request_timeout(3):
    message.reply(...)
```
It can be implemented with using the `contextvars`.

What do you think about that?

JrooTJunior:
I will cancel that issue and start to implement different method of using timeouts.

PR #112, opened 2019-03-22
Some changes

gabbhack:
- Add more await in RedisStorage and RedisStorage2
- Replace "send" with "reply" in some method's names
- Add "answer" method to Message
- Add "lazy_gettext" method to I18nMiddleware
- Reformat code

### DIFF
--- a/aiogram/dispatcher/handler.py
+++ b/aiogram/dispatcher/handler.py
@@ -126,4 +126,4 @@ class Handler:
     class HandlerObj:
         handler: callable
         spec: inspect.FullArgSpec
-        filters: Union[list, tuple, set] = None
+        filters: Union[list, tuple, set, None] = None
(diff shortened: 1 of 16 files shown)


aiogram/dispatcher/handler.py
```diff
@@ -30,6 +34,7 @@ def __init__(self, dispatcher, once=True, middleware_key=None):
         self.once = once
 
         self.handlers = []
+        self.specs = {}
```
JrooTJunior: Why do you place specs to dictionary instead of extending items in `self.handlers`?

aiogram/types/message.py
```diff
@@ -275,6 +276,26 @@ def link(self, text, as_html=True) -> str:
             return md.hlink(text, url)
         return md.link(text, url)
 
+    async def answer(self, text, parse_mode=None, disable_web_page_preview=None,
```
JrooTJunior: What about `answer_photo`, `answer_document` and etc?

PR #119, opened 2019-04-06
Optimize filter checking

gabbhack:


aiogram/dispatcher/filters/__init__.py
```diff
@@ -23,6 +24,8 @@
     'Regexp',
     'StateFilter',
     'Text',
-    'check_filter',
+    'get_filters_spec',
+    'get_filters_spec',
```
JrooTJunior: Line duplicated

PR #122, opened 2019-04-13
Update message.py

rasongame:
I added a type that returns when get_args(), and now you can write something like "Text placed after command"

### DIFF
--- a/aiogram/types/message.py
+++ b/aiogram/types/message.py
@@ -179,7 +179,7 @@ class Message(base.TelegramObject):
         """
         Get arguments
 
-        :return:
+        :return: str
         """
         command = self.get_full_command()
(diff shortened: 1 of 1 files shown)


aiogram/types/message.py
```diff
@@ -179,7 +179,7 @@ def get_args(self):
         """
```
JrooTJunior: add `-> str` to 178 line

aiogram/types/message.py
```diff
@@ -179,7 +179,7 @@ def get_args(self):
...
         Get arguments
 
-        :return:
+        :return: str
```
JrooTJunior: And then here you can write something like "Text placed after command"

PR #153, opened 2019-07-08
Mongo storage

Olegt0rr:
# Description
Added Mongo storage support based on [aiomongo](https://github.com/ZeoAlliance/aiomongo)

### DIFF
--- a/aiogram/contrib/fsm_storage/mongo.py
+++ b/aiogram/contrib/fsm_storage/mongo.py
@@ -1,0 +1,201 @@
+"""
+This module has mongo storage for finite-state machine
+    based on `aiomongo <https://github.com/ZeoAlliance/aiomongo`_ driver
+"""
+
+from typing import Union, Dict, Optional, List, Tuple, AnyStr
+
+import aiomongo
+from aiomongo import AioMongoClient, Database
+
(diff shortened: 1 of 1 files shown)


Birdi7:
Last commit is 2 year ago at [aiomongo](https://github.com/ZeoAlliance/aiomongo). Is this fine?

JrooTJunior:
Is [Motor](https://motor.readthedocs.io/en/stable/) has the same interface? Maybe this storage can be maked more universally for usage with different drivers. Or maybe can be splitted into different storages like `MotorsMongoStorage` and `AioMongoStorage`.

Because Motor looks more reliable

Olegt0rr:
> Last commit is 2 year ago at [aiomongo](https://github.com/ZeoAlliance/aiomongo). Is this fine?

Not aiomongo works fine and it's enought for me.
I'm not sure about motor speed. 2 years ago aiomongo was more faster than motor (look at the benchmarks).

If you can make fresh comparison of motor and aiomongo - please, do it.

JrooTJunior:
I'll merge this PR but after that i rename module from mongo to aiomongo and after that add the motor version of this storage because Motor is official library and is more popular

PR #168, opened 2019-07-18
Add enable_cache parameter to lazy_gettext

Birdi7:
# Description

Add new parameter `enable_cache` to `i18nMiddleware.lazy_gettext`. Specifies if the value is cached after the first call of .value of LazyProxy object. 

## Type of change

# How Has This Been Tested?

Manually

# Checklist:

### DIFF
--- a/aiogram/contrib/middlewares/i18n.py
+++ b/aiogram/contrib/middlewares/i18n.py
@@ -107,7 +107,7 @@ class I18nMiddleware(BaseMiddleware):
         else:
             return translator.ngettext(singular, plural, n)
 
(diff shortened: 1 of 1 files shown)


Birdi7:
Why this parameter is needed:
   When you have a LazyProxy object, after the first calling of .value, the result will be cached, and for the all next invokations, .value will return the same value. 
[src](https://github.com/python-babel/babel/blob/38dfc8dc87f869d4837cbe8338db53ae9e293f8f/babel/support.py#L181).
As a result, if the context locale has changed after the first calling, the value won't be changed

Issue #151, opened 2019-07-08
New filters in builtin module


JrooTJunior:
Next list of filters is planned for implementation in [builtin](https://github.com/aiogram/aiogram/blob/dev-2.x/aiogram/dispatcher/filters/builtin.py) module and registered in filters factory:
- [x] Chat id  ([aiogram#1544](https://t.me/aiogram/1544)) - PR: #172 
- [x] User id ([aiogram#1544](https://t.me/aiogram/1544)) - PR: #172 
- [x] User is admin in chat ([aiogram#1549](https://t.me/aiogram/1549), [aiogram#1555](https://t.me/aiogram/1555)) - PR: #182 
- [x] Message is reply to another message ([aiogram#1549](https://t.me/aiogram/1549)) - PR: #210, #212
- [x] Message forwarded ([aiogram#1549](https://t.me/aiogram/1549)) - PR: #355
- [x] Move [chat type filters](https://github.com/aiogram/aiogram/blob/dev-2.x/aiogram/types/chat.py#L431-L479) ([aiogram#1549](https://t.me/aiogram/1549))
- [x] Multiple text filter ([aiogram_ru#73126](https://t.me/aiogram_ru/73126))
- [x] check that the contact matches the sender - [commit](https://github.com/aiogram/aiogram/commit/5db726d7585c5252343642f1201c4775ac47bfeb)

Birdi7:
I would like to work on Multiple text filter. But I have a question regarding the key word. Should it be `text` for both single and muliple elements, `text` for single and `texts` for a list, or `texts` for both single and multiple elements(like it's done in commands filter)? @JrooTJunior what do you think about it?

Forevka69:
Maybe it will be useful:

Checking if message from group
```
class IsGroup(Filter):
    """
        Check if chat is group
    """

    async def check(self, obj: Union[Message, CallbackQuery], *args):
        if isinstance(obj, Message):
            return obj.chat.id < 0
        return obj.message.chat.id < 0
```

Check if message from private

```
class IsPrivate(Filter):
    async def check(self, obj: Union[Message, CallbackQuery], *args):
        return obj.chat.id == obj.from_user.id
```

Check message text in list 

```
class IsTextInList(Filter):
    def __init__(self, text_list):
        self.text_list = text_list

    async def check(self, obj: Union[Message, CallbackQuery], *args):
        return obj.text in self.text_list
```

JrooTJunior:
[...]

PR #176, opened 2019-07-27
Add multiple text filter

Birdi7:
# Description

Add the multiple text filter as required at #151, `Multiple text filter (aiogram_ru#73126)`.
The following patterns are applied for comparison methods: 
- equals - **True** if **any** of strings in the list is equal to the text
- contains - **True** if the text contains **all** strings from the list. Reason: There is no way to do such comparison right now (except using custom filters), while the condition "if the text contains any string" can be implemented with a bunch of `register_message_handler`
- startswith - **True** if text starts with **any** string from the list. Reason: There is no need to implement "if text starts with all strings from the list" as all prefixes can be mixed in one 
- endswith - **True** if text ends with **any** string from the list. Reason: There is no need to implement "if text ends with all strings from the list" as all postfixes can be mixed in one

Tests are included for all methods. An example of usage was added.

## Type of change

# Checklist:

### DIFF
--- a/aiogram/dispatcher/filters/builtin.py
+++ b/aiogram/dispatcher/filters/builtin.py
@@ -213,11 +213,12 @@ class Text(Filter):
                  ignore_case=False):
         """
         Check text for one of pattern. Only one mode can be used in one filter.
+        In every pattern, a single string is treated as a list with 1 element.
 
-        :param equals:
-        :param contains:
-        :param startswith:
-        :param endswith:
+        :param equals: True if object text in the list
+        :param contains: True if object text contains all strings from the list
+        :param startswith: True if object text startswith any of strings from the list
+        :param endswith: True if object text endswith any of strings from the list
         :param ignore_case: case insensitive
         """
         # Only one mode can be used. check it.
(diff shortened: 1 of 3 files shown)


surik00:
Your **ANY** filters are redundant because the same result can be achieved by adding multiple registrators

Maybe only the _contains all of the entries_ has its place

You have the same code in your example


```python
@dp.message_handler(text_contains='spam')
@dp.message_handler(text_contains='eggs')
await def my_handler(message):
    await foobar()
```

Birdi7:
@surik00 yes, they may be replaced with a bunch of registrations. Although, my filters make code clearer and more readable while don't break the existing code written using previous filters. Command filter works the same way, and it take place in aiogram

JrooTJunior:
I'll uncheck documentation mark until you don't update the documentations.

Birdi7:
@JrooTJunior I added the docs in the last 2 commits. [Here](https://aiogram-birdi7.readthedocs.io/en/add-multiple-text-filter/dispatcher/filters.html#aiogram.dispatcher.filters.builtin.Text) is the built version of docs for this branch

PR #180, opened 2019-07-29
Wrapped function can be registered as handler

### LINKED ISSUE #179: Wrapped coroutine can't be a callback
## Context

Any wrapped coroutine can't be a callback for an event. Instead, the unwrapped coroutine is being registered as handler.

* Operating System: Linux Mint 19.1
* Python Version: 3.7.4
* aiogram version: 2.2
* aiohttp version: 3.5.4
* uvloop version (if installed): 0.12.2

## Expected Behavior

Any wrapped coroutine can be registered as a callback for an event. 

## Current Behavior

The wrapped coroutine isn't being invoked by aiogram, instead, the initial coroutine is being launched.

The failure happens because of unwrapping [here](https://github.com/aiogram/aiogram/blob/f77b752c60180469c55058e7c036d0b0d6598c4b/aiogram/dispatcher/handler.py#L25). This coroutine is being invoked [here](https://github.com/aiogram/aiogram/blob/f77b752c60180469c55058e7c036d0b0d6598c4b/aiogram/dispatcher/handler.py#L59) to get the specifications and the function which will be registered as the handler in the HandlerObj

### Steps to Reproduce

Here is the [listing](https://gist.github.com/Birdi7/4b061f64e62205f63c61472d9171bf1d).
The expected behavior is the printing of "in wrapped" after sending `/start` to a bot, but it doesn't happen because the actual handler is the original coroutine

### How to fix 

This bug can be fixed by returning [here](https://github.com/aiogram/aiogram/blob/f77b752c60180469c55058e7c036d0b0d6598c4b/aiogram/dispatcher/handler.py#L30) the original function instead of unwrapped

Birdi7:
# Description

The wrapped functions can be handler now. AFAIK it won't break anything

Fixes #179 

## Type of change

# Checklist:

### DIFF
--- a/aiogram/dispatcher/handler.py
+++ b/aiogram/dispatcher/handler.py
@@ -23,11 +23,10 @@ class CancelHandler(Exception):
 
 
 def _get_spec(func: callable):
-    wrapped_function = func
     while hasattr(func, '__wrapped__'):  # Try to resolve decorated callbacks
         func = func.__wrapped__
     spec = inspect.getfullargspec(func)
-    return spec, wrapped_function
+    return spec
 
 
 def _check_spec(spec: inspect.FullArgSpec, kwargs: dict):
@@ -56,7 +55,7 @@ class Handler:
         :param filters: list of filters
         :param index: you can reorder handlers
         """
-        spec, handler = _get_spec(handler)
+        spec = _get_spec(handler)
 
         if filters and not isinstance(filters, (list, tuple, set)):
             filters = [filters]
--- a/aiogram/dispatcher/handler.py
+++ b/aiogram/dispatcher/handler.py
@@ -23,11 +23,11 @@ class CancelHandler(Exception):
 
 
 def _get_spec(func: callable):
+    wrapped_function = func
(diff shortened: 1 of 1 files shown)


aiogram/dispatcher/handler.py
```diff
@@ -23,11 +23,11 @@ class CancelHandler(Exception):
 
 
 def _get_spec(func: callable):
+    wrapped_function = func
     while hasattr(func, '__wrapped__'):  # Try to resolve decorated callbacks
         func = func.__wrapped__
-
     spec = inspect.getfullargspec(func)
-    return spec, func
+    return spec, wrapped_function
```
JrooTJunior: In this case maybe don't need to return the handler from this function because decorated handler is already accessible in place from which _get_spec is called

Birdi7: Yes, thank you! Didn't think about it, that's true. Updated in [d2d4928](https://github.com/aiogram/aiogram/pull/180/commits/d2d49282f55544886a38a577d6692c0d7349115c)

Issue #200, opened 2019-08-28
Future does not have _set_result method


apatrushev:
## Context

Future created here:
https://github.com/aiogram/aiogram/blob/58a8f43e2ccc40bce384c18349c9aa3ffe24bd8f/aiogram/dispatcher/dispatcher.py#L80

... later used to _set_result, but asyncio.Future does not have such method:
https://github.com/aiogram/aiogram/blob/58a8f43e2ccc40bce384c18349c9aa3ffe24bd8f/aiogram/dispatcher/dispatcher.py#L302

PR #201, opened 2019-08-28
process cancellation in dispatch polling

### LINKED ISSUE #199: Dispatcher polling task does not process cancellation
## Context

Bare except hides CancelledError:

https://github.com/aiogram/aiogram/blob/58a8f43e2ccc40bce384c18349c9aa3ffe24bd8f/aiogram/dispatcher/dispatcher.py#L287-L290

## Expected Behavior

Exit the loop on CancelledError

apatrushev:
# Description

Exit dispatcher polling loop on CancelledError, fixes #199 

## Type of change

- [v] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/aiogram/dispatcher/dispatcher.py
+++ b/aiogram/dispatcher/dispatcher.py
@@ -284,6 +284,8 @@ class Dispatcher(DataMixin, ContextInstanceMixin):
                 try:
                     with self.bot.request_timeout(request_timeout):
                         updates = await self.bot.get_updates(limit=limit, offset=offset, timeout=timeout)
+                except asyncio.CancelledError:
+                    break
                 except:
                     log.exception('Cause exception while getting updates.')
                     await asyncio.sleep(error_sleep)


JrooTJunior:
As the same of #200 need to change here `self._close_waiter._set_result(None)` to `self._close_waiter.set_result(None)` in finally block.

apatrushev:
> As the same of #200 need to change here `self._close_waiter._set_result(None)` to `self._close_waiter.set_result(None)` in finally block.

Sorry, I did not understand what do you mean. It is different issue which fixes proper task stopping by exiting the loop on CancelledError (standard way to stop async io tasks). https://docs.python.org/3/library/asyncio-exceptions.html

JrooTJunior:
I say about next code block:
https://github.com/aiogram/aiogram/blob/a4f684f6bf6a1166fc98552e4a134f400b0995fd/aiogram/dispatcher/dispatcher.py#L303-L305

But.. Okay. I will change this code.

PR #210, opened 2019-09-16
add IsReplyFilter

Arwichok:
# Description

Add IsReplyFilter

# Use
```py
@dp.message_handler(is_reply=True)
async def reply_test(msg, reply):
    await msg.answer(f'Reply: {reply.text}')
```

### DIFF
--- a/aiogram/dispatcher/filters/builtin.py
+++ b/aiogram/dispatcher/filters/builtin.py
@@ -637,5 +637,7 @@ class IsReplyFilter(BoundFilter):
         self.is_reply = is_reply
 
     async def check(self, msg: Message):
-        if msg.reply_to_message:
+        if msg.reply_to_message and self.is_reply:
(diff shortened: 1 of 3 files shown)


aiogram/dispatcher/filters/builtin.py
```diff
@@ -625,3 +625,19 @@ def validate(cls, full_config: typing.Dict[str, typing.Any]) -> typing.Optional[
...
+    """
+    key = 'is_reply'
+
+    def __init__(self, is_reply):
+        self.is_reply = is_reply
+
+    async def check(self, msg: Message):
+        if msg.reply_to_message and self.is_reply:
+            return {'reply': msg.reply_to_message}
+        elif msg.reply_to_message is None and self.is_reply is False:
```
JrooTJunior:
```suggestion
        elif not msg.reply_to_message and not self.is_reply:
```

PR #218, opened 2019-10-12
Update message.py

Bunk100:
Updated send_copy:
Added the ability to specify reply_markup, parse_mode

### DIFF
--- a/aiogram/types/message.py
+++ b/aiogram/types/message.py
@@ -1596,7 +1596,7 @@ class Message(base.TelegramObject):
             if parse_mode == 'markdown':
                 text = self.md_text if (self.text or self.caption) else None
         if reply_markup:
-            kwargs["reply_markup"] = self.reply_markup
+            kwargs["reply_markup"] = reply_markup
 
         if self.text:
             return await self.bot.send_message(text=text, **kwargs)
--- a/aiogram/types/message.py
+++ b/aiogram/types/message.py
@@ -1567,29 +1567,36 @@ class Message(base.TelegramObject):
     async def send_copy(
             self: Message,
             chat_id: typing.Union[str, int],
-            with_markup: bool = False,
             disable_notification: typing.Optional[bool] = None,
             reply_to_message_id: typing.Optional[int] = None,
+            reply_markup: typing.Union[InlineKeyboardMarkup, ReplyKeyboardMarkup, None] = self.reply_markup,
+            parse_mode: typing.Union[base.String, None] = None,
     ) -> Message:
         """
         Send copy of current message
 
         :param chat_id:
-        :param with_markup:
         :param disable_notification:
         :param reply_to_message_id:
+        :param reply_markup:
+        :param parse_mode:
         :return:
         """
-        kwargs = {"chat_id": chat_id, "parse_mode": ParseMode.HTML}
+        kwargs = {"chat_id": chat_id}
+        text = self.text or self.caption
 
         if disable_notification is not None:
             kwargs["disable_notification"] = disable_notification
         if reply_to_message_id is not None:
             kwargs["reply_to_message_id"] = reply_to_message_id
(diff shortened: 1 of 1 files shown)


aiogram/types/message.py
```diff
@@ -1567,30 +1567,37 @@ def link(self, text, as_html=True) -> str:
     async def send_copy(
             self: Message,
             chat_id: typing.Union[str, int],
-            with_markup: bool = False,
             disable_notification: typing.Optional[bool] = None,
             reply_to_message_id: typing.Optional[int] = None,
+            reply_markup: typing.Union[InlineKeyboardMarkup, ReplyKeyboardMarkup, None] = self.reply_markup,
```
gabbhack: It's doesn't work

aiogram/types/message.py
```diff
@@ -1567,29 +1567,36 @@ def link(self, text, as_html=True) -> str:
...
         :param chat_id:
-        :param with_markup:
         :param disable_notification:
         :param reply_to_message_id:
+        :param reply_markup:
+        :param parse_mode:
         :return:
         """
-        kwargs = {"chat_id": chat_id, "parse_mode": ParseMode.HTML}
+        kwargs = {"chat_id": chat_id, "reply_markup": self.reply_markup}
```
gabbhack: Can we use `{"reply_markup": reply_markup or self.reply_markup}`?

aiogram/types/message.py
```diff
@@ -1567,29 +1567,36 @@ def link(self, text, as_html=True) -> str:
     async def send_copy(
             self: Message,
             chat_id: typing.Union[str, int],
-            with_markup: bool = False,
             disable_notification: typing.Optional[bool] = None,
             reply_to_message_id: typing.Optional[int] = None,
+            reply_markup: typing.Union[InlineKeyboardMarkup, ReplyKeyboardMarkup, None] = None,
+            parse_mode: typing.Union[base.String, None] = None,
```
gabbhack: I think parse_mode is not needed, we can use HTML by default.

aiogram/types/message.py
Bunk100: Hmm, I think this is better than using if

Bunk100: > Can we use `{"reply_markup": reply_markup or self.reply_markup}`?

Hmm, I think this is better than using if

aiogram/types/message.py
```diff
@@ -1582,19 +1582,15 @@ def link(self, text, as_html=True) -> str:
...
             kwargs["disable_notification"] = disable_notification
         if reply_to_message_id is not None:
             kwargs["reply_to_message_id"] = reply_to_message_id
-        if parse_mode is not None:
-            kwargs["parse_mode"] = parse_mode
-            if parse_mode == 'html':
-                text = self.html_text if (self.text or self.caption) else None
-            if parse_mode == 'markdown':
-                text = self.md_text if (self.text or self.caption) else None
+        if not kwargs.get("reply_markup"):
```
gabbhack: I think it is not necessary, since we already [have a check](https://github.com/aiogram/aiogram/blob/dev-2.x/aiogram/utils/payload.py#L26) for None.

aiogram/types/message.py
```diff
@@ -1567,29 +1567,30 @@ def link(self, text, as_html=True) -> str:
...
+                    "chat_id": chat_id, 
+                    "reply_markup": reply_markup or self.reply_markup,
+                    "parse_mode": ParseMode.HTML
+                 }
+        text = self.html_text if (self.text or self.caption) else None
 
         if disable_notification is not None:
             kwargs["disable_notification"] = disable_notification
         if reply_to_message_id is not None:
             kwargs["reply_to_message_id"] = reply_to_message_id
```
Piterden: Ну задай весь объект через {}. Ведь у каждого параметра есть значение по умолчанию.

PR #231, opened 2019-11-11
WIP: Explain how add new custom route for Application which use for webhook

Forevka:
# Description

Explain how to add new custom route for Application which use for webhook

Fixes # (issue)

## Type of change

**Test Configuration**:
* Operating System: win 10
* Python version: 3.7.1

# Checklist:

### DIFF
--- a/aiogram/__init__.py
+++ b/aiogram/__init__.py
@@ -38,5 +38,5 @@ __all__ = [
     'utils'
 ]
 
-__version__ = '2.3.dev1'
+__version__ = '2.4'
 __api_version__ = '4.4'
--- a/examples/custom_route.py
+++ b/examples/custom_route.py
@@ -50,8 +50,8 @@ class CustomExecutor(Executor):
         inherit from Executor class and overwrite run_app
         method so you can easily add own routes
     '''
-    def run_app(self, **kwargs):
-        web.run_app(self._web_app, **kwargs, port = 9999)
+    def run_app(self, port: int = 8080, **kwargs,):
+        web.run_app(self._web_app, **kwargs, port = port)
     
     def add_subapp(self, subapp: web.Application, subapp_prefix: str):
         self._web_app.add_subapp(subapp_prefix, subapp)
--- a/examples/custom_route.py
+++ b/examples/custom_route.py
@@ -1,0 +1,82 @@
+import logging
+from aiohttp import web
+import typing
+
+from aiogram import Bot, Dispatcher, executor, types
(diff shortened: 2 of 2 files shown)


examples/custom_route.py
```diff
@@ -0,0 +1,82 @@
...
+from aiogram import Bot, Dispatcher, executor, types
+from aiogram.dispatcher.webhook import WebhookRequestHandler
+from aiogram.utils.executor import Executor
+from loguru import logger
+
+
+API_TOKEN = 'BOT TOKEN'
+
+# Configure logging
+logging.basicConfig(level=logging.INFO)
```
uwinx, line 14: why did you import `loguru` above?

examples/custom_route.py
```diff
@@ -0,0 +1,82 @@
...
+async def who_am_i(request: web.Request):
+    '''
+        Return bot info from this route
+    '''
+    me = await request.app['BOT_DISPATCHER'].bot.me
+    return web.json_response(me.as_json())
+
+
+
+class CustomExecutor(Executor):
```
uwinx, line 47: Why exactly are we inheriting from the executor class? Is it to wrap `_web_app` initialized in parent executor?

aiogram/__init__.py
```diff
@@ -38,5 +38,5 @@
     'utils'
 ]
 
-__version__ = '2.3.dev1'
+__version__ = '2.4'
```
uwinx, line 41: Examples do not require version bumping, and if your example depends on specific version, you can note that in example module's docstring.

Forevka: i`m going to remove this. Dont remember why i do this

examples/custom_route.py
Forevka: it was only for test purpose. You right i should remove this line it may confuse.

examples/custom_route.py
Forevka: Yes, its for work directly with _web_app instance

uwinx: I don't think this is a good idea. Why don't you work with `aiohttp.web.Application` directly? It's little unclear as for me and would not be very obvious for beginners.

uwinx:
@Forevka, Maybe you could reconsider the short description to this example? Something like "Integrating bot with an existing aiohttp aiohttp app" would sound nice, wouldn't it?

uwinx:
@Forevka, how's the progress so far?

uwinx:
[aiogram chat](https://t.me/aiogram_ru/212911) (ru). Given that, I'm closing PR.

PR #240, opened 2019-11-23
#239 check_token fix

### LINKED ISSUE #239: check_token bug report
## Context

```python
from aiogram.bot.api import check_token

check_token('1234:') # True
check_token('12:ABC-DEF1234ghIkl-zyx57W2v1u123ew11') # ValidationError
```
* aiogram version: 2.4

## Expected Behavior

In the first case I expect an exception, in the second I expect True

Olegt0rr:
# Description

Added check: right part of token is exists.
Removed check: left part of token is longer than 3 symbols
Added test cases from issue.
Also added token type validation and test cases for it.

Fixes #239 
Fixes token type validation (not issued)

## Type of change

## How Has This Been Tested?
- [ ] TestCheckToken

### DIFF
--- a/aiogram/bot/api.py
+++ b/aiogram/bot/api.py
@@ -28,7 +28,7 @@ def check_token(token: str) -> bool:
         raise exceptions.ValidationError('Token is invalid!')
 
     left, sep, right = token.partition(':')
-    if (not sep) or (not left.isdigit()) or (len(left) < 3):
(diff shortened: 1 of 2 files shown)


Olegt0rr:
Also added token type validation and test cases for it.
For example user can get None from env and pass it to bot as token.

Then we get:
```python
if any(x.isspace() for x in token):
TypeError: 'NoneType' object is not iterable
```

PR #259, opened 2020-01-12
Fix code style

ResQNetM:
Removed unnecessary f-strings

### DIFF
--- a/aiogram/dispatcher/handler.py
+++ b/aiogram/dispatcher/handler.py
@@ -44,6 +44,10 @@ class Handler:
         self.handlers: List[Handler.HandlerObj] = []
         self.middleware_key = middleware_key
 
+    @property
+    def formatted_middleware_key(self):
+        return "pre_process_"+self.middleware_key
+
     def register(self, handler, filters=None, index=None):
         """
         Register callback
@@ -98,7 +102,7 @@ class Handler:
 
         if self.middleware_key:
             try:
-                await self.dispatcher.middleware.trigger(f"pre_process_{self.middleware_key}", args + (data,))
+                await self.dispatcher.middleware.trigger(self.formatted_middleware_key, args + (data,))
             except CancelHandler:  # Allow to cancel current event
                 return results
 
@@ -112,7 +116,7 @@ class Handler:
                     ctx_token = current_handler.set(handler_obj.handler)
                     try:
                         if self.middleware_key:
-                            await self.dispatcher.middleware.trigger(f"process_{self.middleware_key}", args + (data,))
+                            await self.dispatcher.middleware.trigger(self.formatted_middleware_key, args + (data,))
                         partial_data = _check_spec(handler_obj.spec, data)
                         response = await handler_obj.handler(*args, **partial_data)
                         if response is not None:
@@ -127,7 +131,7 @@ class Handler:
                         current_handler.reset(ctx_token)
         finally:
             if self.middleware_key:
-                await self.dispatcher.middleware.trigger(f"post_process_{self.middleware_key}",
+                await self.dispatcher.middleware.trigger(f"post_process_"+self.middleware_key,
                                                          args + (results, data,))
 
         return results
--- a/aiogram/dispatcher/handler.py
+++ b/aiogram/dispatcher/handler.py
@@ -44,10 +44,6 @@ class Handler:
         self.handlers: List[Handler.HandlerObj] = []
         self.middleware_key = middleware_key
 
-    @property
-    def formatted_middleware_key(self):
-        return "pre_process_"+self.middleware_key
-
     def register(self, handler, filters=None, index=None):
         """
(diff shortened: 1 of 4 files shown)


aiogram/dispatcher/middlewares.py
```diff
@@ -35,7 +35,7 @@ def setup(self, middleware):
 
         self.applications.append(middleware)
         middleware.setup(self)
-        log.debug(f"Loaded middleware '{middleware.__class__.__name__}'")
+        log.debug(f"Loaded middleware "+repr(middleware.__class__.__name__))
```
JrooTJunior: For logging is recommended to use %-style formating because when log level is disabled formatter will never be called.
```suggestion
        log.debug("Loaded middleware %r", middleware.__class__.__name__)
```

aiogram/dispatcher/handler.py
```diff
@@ -112,7 +116,7 @@ def unregister(self, handler):
                     ctx_token = current_handler.set(handler_obj.handler)
                     try:
                         if self.middleware_key:
-                            await self.dispatcher.middleware.trigger(f"process_{self.middleware_key}", args + (data,))
+                            await self.dispatcher.middleware.trigger(self.formatted_middleware_key, args + (data,))
```
JrooTJunior: `self.formatted_middleware_key` can't be used here because `process_<key>` is expected instead of `pre_process_<key>`

aiogram/dispatcher/handler.py
```diff
@@ -44,6 +44,10 @@ def __init__(self, dispatcher, once=True, middleware_key=None):
         self.handlers: List[Handler.HandlerObj] = []
         self.middleware_key = middleware_key
 
+    @property
+    def formatted_middleware_key(self):
+        return "pre_process_"+self.middleware_key
```
JrooTJunior: middleware key is Optional[str]. in this case simple concatenation can make an TypeError
f-string is more safe in this case.

Also this key is used in single place it makes no sense to separately take out this attribute. Use in-place formatting instead.

aiogram/dispatcher/storage.py
```diff
@@ -444,7 +444,7 @@ def __contains__(self, item):
         return item in self._data
 
     def __str__(self):
-        readable_state = f"'{self.state}'" if self.state else "<default>"
+        readable_state = repr(str(self.state)) if self.state else "<default>"
```
JrooTJunior: Converting to string is not needed here because self.state is Optional[str] and case with None already covered by this ternary-operator.
```suggestion
        readable_state = repr(self.state) if self.state else "<default>"
```

aiogram/dispatcher/webhook.py
```diff
@@ -261,7 +261,7 @@ def validate_ip(self):
         if self.request.app.get('_check_ip', False):
             ip_address, accept = self.check_ip()
             if not accept:
-                log.warning(f"Blocking request from an unauthorized IP: {ip_address}")
+                log.warning(f"Blocking request from an unauthorized IP: "+ip_address)
```
JrooTJunior: `ip_address` can be `None`. Best way in this case - use %-style (the same with my previous comment for logging)
```suggestion
                log.warning(f"Blocking request from an unauthorized IP: %s", ip_address)
```

aiogram/dispatcher/handler.py
```diff
@@ -98,7 +102,7 @@ def unregister(self, handler):
 
         if self.middleware_key:
             try:
-                await self.dispatcher.middleware.trigger(f"pre_process_{self.middleware_key}", args + (data,))
+                await self.dispatcher.middleware.trigger(self.formatted_middleware_key, args + (data,))
```
JrooTJunior: Don't change this place.
This change has no positive effect.

aiogram/dispatcher/handler.py
```diff
@@ -127,7 +131,7 @@ def unregister(self, handler):
                         current_handler.reset(ctx_token)
         finally:
             if self.middleware_key:
-                await self.dispatcher.middleware.trigger(f"post_process_{self.middleware_key}",
+                await self.dispatcher.middleware.trigger(f"post_process_"+self.middleware_key,
```
JrooTJunior: The same with previous cases. Does not make sense.

aiogram/dispatcher/middlewares.py
```diff
@@ -96,7 +96,7 @@ def is_configured(self) -> bool:
         :param args:
         :return:
         """
-        handler_name = f"on_{action}"
+        handler_name = f"on_"+action
```
JrooTJunior: Doesn't follow project code-style. plus should be separated by space with left and right parts of expression.

aiogram/dispatcher/handler.py
ResQNetM: but what's the point of a f-string here

PR #268, opened 2020-01-29
Add GitHub Actions workflow to publish packages

### LINKED ISSUE #235: Automatically build and deploy AUR package on new release
This issue is just for me, I'll implement this feature and open PR some time.

evgfilim1:
## Description
Add GitHub Actions workflow to publish packages to [PyPI][pypi] and [AUR][aur] (Fixes #235)

To make this work, you must create these secrets:
- `PYPI_TOKEN` — PyPI token which is used to publish packages
- `AUR_SSH_KEY` — **private** SSH key configured to access AUR
- `GIT_EMAIL` — email of git committer (will be shown in AUR-git repo)
- `GIT_NAME` — name of git committer (will be shown in AUR-git repo)

## Type of change
- New feature (non-breaking change which adds functionality)

## How has this been tested?
It was tested with some changes in my fork with GitHub Actions, but _this_ version wasn't tested as it requires aiogram 3.x to be released.

[pypi]: https://pypi.org/project/aiogram/
[aur]: https://aur.archlinux.org/packages/python-aiogram/

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,48 @@
+name: Publish packages
+
+on:
+    push:
+        branches:
+          - 'releases/**'
+        tags:
+          - v3
+          - v3.*
+
+jobs:
+    publish:
+        runs-on: ubuntu-latest
+        steps:
+          - name: Checkout repo
+            uses: actions/checkout@v2
+          - name: Set up Python 3.7
+            uses: actions/setup-python@v1
+            with:
+                python-version: 3.7
+          - name: Install dependencies
+            run: |
+                python -m pip install --upgrade pip poetry==1.0
+          - name: Build wheel
+            run: |
+                poetry build
+          - name: Publish to PyPI
+            env:
+                POETRY_PYPI_TOKEN_PYPI: ${{ secrets.PYPI_TOKEN }}
+            run: |
+                poetry publish
+          - name: Prepare AUR package
+            env:
+                AUR_SSH_KEY: ${{ secrets.AUR_SSH_KEY }}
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,48 @@
...
+          - name: Set up Python 3.7
+            uses: actions/setup-python@v1
+            with:
+                python-version: 3.7
+          - name: Install dependencies
+            run: |
+                python -m pip install --upgrade pip poetry==1.0
+          - name: Build wheel
+            run: |
+                poetry build
```
b0g3r: What about additional step, which will check that version from pyproject.toml is equal to git tag?

JrooTJunior: Sounds good
And `__version__` from `aiogram/__init__.py` is also should be validated

b0g3r: Should we use a tool like `bump2version` or `dephell.versioning` for reducing manual work?

evgfilim1: @b0g3r check and.. what's next?

b0g3r: If it's not we have to stop and mark the action as failed I think

It's like 'oh uh you forgot to bump version in pyproject.toml and __init__.py would you please bump it in before releasing?'

b0g3r: In such cases re-pushing the git tag is dirty but acceptable workaround

evgfilim1:
I've lost access to source repo (deleted my fork), so I'll recreate this PR later with all needed fixes and improvements.

PR #282, opened 2020-03-16
WIP: Add download_file and URLInputFile

gabbhack:
# Description

- Add new download_file method to Bot
- Add new URLInputFile type

## Type of change

# Checklist:

### DIFF
--- a/aiogram/api/types/mixins.py
+++ b/aiogram/api/types/mixins.py
@@ -1,0 +1,89 @@
+import os
+import pathlib
+from typing import BinaryIO, Optional, Union
+
+from aiogram.api.types.file import File
+
+
+class Downloadable:
+    """
+    Mixin for files
+    """
+    async def download(
+        self,
+        destination: Optional[Union[BinaryIO, pathlib.Path, str]] = None,
+        timeout: int = 30,
+        chunk_size: int = 65536,
+        seek: bool = True,
+        make_dirs: bool = True,
+    ) -> Optional[BinaryIO]:
+        """
+        Download file
+
+        :param destination: Filename, file path or instance of :class:`io.IOBase`. For e.g. :class:`io.BytesIO`, defaults to None
+        :type destination: Optional[Union[BinaryIO, pathlib.Path, str]]
+        :param timeout: Total timeout in seconds, defaults to 30
+        :type timeout: int
+        :param chunk_size: File chunks size, defaults to 64 kb
+        :type chunk_size: int
+        :param seek: Go to start of file when downloading is finished. Used only for destination with :class:`typing.BinaryIO` type, defaults to True
+        :type seek: bool
+        :param make_dirs: make dirs if not exist
+        :return: destination
+        """
+        from aiogram.api.client.bot import Bot
+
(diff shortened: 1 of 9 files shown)


gabbhack:
I have no idea how to implement Downloadable mixin correctly.

aiogram/api/types/mixins.py
```diff
@@ -0,0 +1,89 @@
...
+        :param timeout: Total timeout in seconds, defaults to 30
+        :type timeout: int
+        :param chunk_size: File chunks size, defaults to 64 kb
+        :type chunk_size: int
+        :param seek: Go to start of file when downloading is finished. Used only for destination with :class:`typing.BinaryIO` type, defaults to True
+        :type seek: bool
+        :param make_dirs: make dirs if not exist
+        :return: destination
+        """
+        from aiogram.api.client.bot import Bot
```
b0g3r: Why local imports everywhere 🤔 

Did I miss something?

b0g3r: From chat: possible circular imports

aiogram/api/types/mixins.py
```diff
@@ -0,0 +1,89 @@
...
+        Attention!!
+        This method has security vulnerabilities for the reason that result
+        contains bot's *access token* in open form. Use at your own risk!
+
+        :return: url
+        """
+        from aiogram.api.client.bot import Bot
+
+        file = await self.get_file()
+        return Bot.get_current(no_error=False).session.api.file_url(file.file_path)
```
b0g3r: You missed the token param
https://github.com/aiogram/aiogram/blob/0bd7fc2c7e7a4e9cb9976428a6345b57ff091a4a/aiogram/api/client/telegram.py#L23

aiogram/api/types/mixins.py
```diff
@@ -0,0 +1,89 @@
...
+        """
+        from aiogram.api.client.bot import Bot
+
+        file = await self.get_file()
+
+        is_path = True
+        if destination is None:
+            destination = file.file_path
+        elif isinstance(destination, (str, pathlib.Path)) and os.path.isdir(destination):
+            # file from self.get_file() always have file_path
```
b0g3r: Why? Docs says File object has optional file_path: https://core.telegram.org/bots/api#file

aiogram/api/types/mixins.py
gabbhack: Yes :(
It looks like we need to add a separate method to the Bot.

aiogram/api/types/mixins.py
gabbhack: This is strange. File can only be obtained from the [getFile](https://core.telegram.org/bots/api#getfile) method, the description of which says "The file can then be downloaded via the link https://api.telegram.org/file/bot<token>/<file_path>, where **<file_path> is taken from the response**".

I thought that `file_path` would be missing for a large file (more than 20MB), but in this case an error is returned "Bad Request: file is too big". So I don't know when file_path can come empty. I asked a question to the technical support of bots.

PR #283, opened 2020-03-18
:label: Turn on mypy's strict mode

b0g3r:


aiogram/dispatcher/event/handler.py
```diff
@@ -19,21 +29,22 @@ class CallableMixin:
...
+    def _prepare_kwargs(self, kwargs: Dict[str, Any]) -> Dict[str, Any]:
         if self.spec.varkw:
             return kwargs
 
         return {k: v for k, v in kwargs.items() if k in self.spec.args}
 
-    async def call(self, *args, **kwargs):
-        wrapped = partial(self.callback, *args, **self._prepare_kwargs(kwargs))
+    async def call(self, *args: Any, **kwargs: Any) -> Any:
+        # TODO: what we should do if callback is BaseHandler?
```
b0g3r: BaseHandler isn't callable so partial can break

aiogram/dispatcher/event/handler.py
```diff
@@ -49,10 +60,10 @@ class HandlerObject(CallableMixin):
     callback: HandlerType
     filters: Optional[List[FilterObject]] = None
 
-    def __post_init__(self):
+    def __post_init__(self) -> None:
         super(HandlerObject, self).__post_init__()
-
-        if inspect.isclass(self.callback) and issubclass(self.callback, BaseHandler):
+        # TODO: by types callback must be Callable or BaseHandler, not Type[BaseHandler]
+        if inspect.isclass(self.callback) and issubclass(self.callback, BaseHandler):  # type: ignore
```
b0g3r: issubclass expect class-object, but self.callback can be BaseHandler instance, not class

Is it error in type?

aiogram/utils/helper.py
```diff
@@ -135,69 +144,70 @@ def apply(cls, text, mode):
...
 
+class _BaseItem:
+    def __init__(self, value: Optional[str] = None):
+        self._value = cast(str, value)
+
+    def __set_name__(self, owner: Any, name: str) -> None:
+        if not name.isupper():
+            raise NameError("Name for helper item must be in uppercase!")
+        if not self._value:
+            # TODO: а если не имеет?
```
b0g3r: What we should do if owner hasn't attribute "mode" and _value is None? 🤔

aiogram/utils/helper.py
```diff
@@ -135,69 +144,70 @@ def apply(cls, text, mode):
             return mode(text)
         return text
 
+class _BaseItem:
+    def __init__(self, value: Optional[str] = None):
+        self._value = cast(str, value)
```
b0g3r: Temporary solution

aiogram/dispatcher/handler/message.py
```diff
@@ -24,5 +27,6 @@ class MessageHandlerCommandMixin(BaseHandlerMixin[Message]):
     @property
     def command(self) -> Optional[CommandObject]:
         if "command" in self.data:
-            return self.data["command"]
+            # TODO: remove cast
+            return cast(CommandObject, self.data["command"])
```
b0g3r: Can we add typing for BaseHandler.data?

aiogram/dispatcher/handler/base.py
```diff
@@ -23,18 +31,20 @@ def __init__(self, event: T, **kwargs: Any) -> None:
...
         if "bot" in self.data:
-            return self.data["bot"]
+            # TODO: remove cast
+            return cast(Bot, self.data["bot"])
         return Bot.get_current()
 
     @property
     def update(self) -> Update:
-        return self.data["update"]
+        # TODO: remove cast
```
b0g3r: Can we add typing for BaseHandler.data?

aiogram/dispatcher/handler/base.py
```diff
@@ -23,18 +31,20 @@ def __init__(self, event: T, **kwargs: Any) -> None:
         self.data: Dict[str, Any] = kwargs
 
     @property
-    def bot(self) -> Bot:
+    def bot(self) -> Optional[Bot]:
         if "bot" in self.data:
-            return self.data["bot"]
+            # TODO: remove cast
```
b0g3r: Can we add typing for BaseHandler.data?

aiogram/dispatcher/handler/base.py
```diff
@@ -23,18 +31,20 @@ def __init__(self, event: T, **kwargs: Any) -> None:
...
+            return cast(Bot, self.data["bot"])
         return Bot.get_current()
```
b0g3r: Maybe it's a good idea to use get_current(no_error=False) there?

aiogram/dispatcher/filters/base.py
```diff
@@ -1,22 +1,26 @@
...
+    pass
 
-        pass
-    else:  # pragma: no cover
 
-        @abstractmethod
-        async def __call__(self, *args: Any, **kwargs: Any) -> Union[bool, Dict[str, Any]]:
-            pass
+class BaseFilter(ABC, BaseModel):
+    # This little hack with typehint is needed because mypy checks validity of overrides and raises:
```
b0g3r: I remove your hack by my own hack 🌚

aiogram/api/client/base.py
```diff
@@ -54,14 +62,15 @@ def id(self) -> int:
...
+    async def context(self, auto_close: bool = True) -> AsyncIterator['BaseBot[ContextInstance]']:
         """
         Generate bot context
 
         :param auto_close:
         :return:
         """
-        token = self.set_current(self)
+        # TODO: because set_current expects Bot, not BaseBot — this check fails
+        token = self.set_current(self)  # type: ignore
```
b0g3r: this method IRL expects be running only via Bot class, not BaseBot class

If we want an honest typing, we should rewrite BaseBot -> Bot relation

aiogram/api/client/bot.py
```diff
@@ -101,12 +101,12 @@
...
 
 
-class Bot(BaseBot):
+class Bot(BaseBot['Bot']):
     """
     Class where located all API methods
     """
 
-    @alru_cache()
+    @alru_cache()  # type: ignore
```
b0g3r: alru_cache doesn't have typing :(

aiogram/api/types/base.py
```diff
@@ -5,7 +5,7 @@
 from aiogram.utils.mixins import ContextInstanceMixin
 
 
-class TelegramObject(ContextInstanceMixin, BaseModel):
+class TelegramObject(ContextInstanceMixin["TelegramObject"], BaseModel):
```
JrooTJunior: In due to [PEP-563](https://www.python.org/dev/peps/pep-0563/) you can remove here brackets in type hint by adding in top of file next line:
```python3
from future import __annotations__
```

And then you can do this:
```suggestion
class TelegramObject(ContextInstanceMixin[TelegramObject], BaseModel):
```

aiogram/dispatcher/event/handler.py
JrooTJunior: What can break?

classes is always callable and instances of BaseHandler is awaitable i don't think partial can break something

aiogram/dispatcher/event/handler.py
JrooTJunior: Is not an error in type because we are registering subclasses of BaseHandler instead of instances of that classes

Example:
```python3
@router.message_handler()
class MyMessageHandler(MessageHandler):
    async def handle(self):
        ...

```

aiogram/dispatcher/filters/base.py
JrooTJunior: k
:rofl:

aiogram/dispatcher/handler/base.py
JrooTJunior: data has type `Dict[str, Any]` and more specific type can not be placed because anybody can add values to this mapping from filters and middlewares.

aiogram/dispatcher/handler/base.py
JrooTJunior: I think yes

aiogram/dispatcher/handler/base.py
b0g3r: So, what's better: use casting or isinstance?

aiogram/api/client/bot.py
b0g3r: No, it isn't solution, because strict mypy has setting about typed decorators and alru_cache has not typing.

aiogram/utils/helper.py
b0g3r: @JrooTJunior

aiogram/api/client/base.py
b0g3r: We should remove BaseBot class and move it into Bot class

aiogram/api/client/bot.py
```diff
@@ -98,15 +111,92 @@
...
+from ...utils.token import (
+    validate_token,
+    extract_bot_id,
+)
+
+T = TypeVar("T")
 
 
-class Bot(BaseBot):
+class Bot(ContextInstanceMixin["Bot"]):
```
JrooTJunior: Just add at top of the file:
```python3
from __future__ import annotations
```
And then do:
```suggestion
class Bot(ContextInstanceMixin[Bot]):
```

tests/test_utils/test_mixins.py
```diff
@@ -1,13 +1,16 @@
...
 
-from aiogram.utils.mixins import ContextInstanceMixin, DataMixin
+from aiogram.utils.mixins import (
+    ContextInstanceMixin,
+    DataMixin,
+)
 
 
-class DataObject(DataMixin):
+class ContextObject(ContextInstanceMixin["ContextObject"]):
```
JrooTJunior: Just use postponed evaluation of annotations (`from __future__ import annotations`)

aiogram/dispatcher/handler/base.py
JrooTJunior: If needed to do different actions on different types - use isisntance(like raising exceptions or converting data and etc.) in other cases need to use casting

aiogram/utils/helper.py
JrooTJunior: Helper is very controversial feature, I do not know what to do with it

JrooTJunior: Woops. It seems like bug in all previous versions of aiogram.

Originally Item was planned to use only as descriptor for Helper classes and in this case owner is always has the `mode` attribute.

I think need to add validation of parent
```python3
if not inspect.isclass(owner) or not issubclass(owner, Helper):
    raise RuntimeError("Item classes can be used only as Helper attributes")
```

aiogram/api/client/bot.py
b0g3r: Nope:
```python
from __future__ import annotations

from typing import (
    TypeVar,
    Generic,
)

T = TypeVar("T")


class Genericable(Generic[T]):
    pass

class Quotes(Genericable[Quotes]):
    pass
```

```bash
>>> python a.py
NameError: name 'Quotes' is not defined
```

aiogram/utils/helper.py
b0g3r: What about OrderedHelper? It isn't a subclass of Helper

b0g3r: Okay, I fixed it with adding Helper as OrderedHelper parent

PR #284, opened 2020-03-19
feat(proxy): proxy for aiohttp,base sessions

uwinx:
# Description

Add support for proxies in `aiohttp` session with `aiohttp_socks` library,
edit `BaseSession` class to support proxies for other sessions in future.

## Type of change

# Checklist:


aiogram/api/client/session/aiohttp.py
```diff
@@ -1,29 +1,72 @@
...
+        )
         self._session: Optional[ClientSession] = None
+        self.cfg.connector_type = TCPConnector
+        self.cfg.connector_init = cast(Dict[str, Any], {})
+
+        if self.proxy:
+            try:
+                from aiohttp_socks import ProxyConnector
+                from aiohttp_socks.utils import parse_proxy_url
+            except ImportError as exc:
```
JrooTJunior:
```suggestion
            except ImportError as exc:  # pragma: no cover
```

aiogram/api/client/session/aiohttp.py
```diff
@@ -1,29 +1,72 @@
...
+            json_loads=json_loads,
+            json_dumps=json_dumps,
+            proxy=proxy
+        )
         self._session: Optional[ClientSession] = None
+        self.cfg.connector_type = TCPConnector
+        self.cfg.connector_init = cast(Dict[str, Any], {})
+
+        if self.proxy:
+            try:
```
JrooTJunior: What about HTTP proxy?
aiohttp already supports HTTP proxy and in this case external library is not needed

aiogram/api/client/session/base.py
```diff
@@ -30,6 +33,9 @@ def __init__(
         self.api = api
         self.json_loads = json_loads
         self.json_dumps = json_dumps
+        self.proxy = proxy
+
+        self.cfg: types.SimpleNamespace = types.SimpleNamespace()
```
JrooTJunior: I don't like short name for this attribute. What do you think about something like `properties` or `config`?

And also i have question why do you use NS instead of adding separated attributes?

pyproject.toml
```diff
@@ -62,6 +62,7 @@ pymdown-extensions = "^6.1"
 lxml = "^4.4"
 ipython = "^7.10"
 markdown-include = "^0.5.1"
+aiohttp-socks = "^0.3.4"
```
JrooTJunior: *aiohttp-socks* is also should be added to `[tool.poetry.dependencies]` block as optional dependency (See `uvloop`) and in block `[tool.poetry.extras]` as `proxy = ["aiohttp-socks"]`

pyproject.toml
```diff
@@ -62,6 +62,7 @@ pymdown-extensions = "^6.1"
...
 
 [tool.poetry.extras]
 fast = ["uvloop"]
```
JrooTJunior:
```suggestion
fast = ["uvloop"]
proxy = ["aiohttp-socks"]
```

aiogram/api/client/session/aiohttp.py
```diff
@@ -1,29 +1,72 @@
...
+                from aiohttp_socks import ProxyConnector
+                from aiohttp_socks.utils import parse_proxy_url
+            except ImportError as exc:
+                raise UserWarning(
+                    "In order to use aiohttp client for proxy requests, install "
+                    "https://pypi.org/project/aiohttp-socks/"
+                ) from exc
+
+            if isinstance(self.proxy, str):
+                proxy_url, proxy_auth = self.proxy, None
```
JrooTJunior: This case should be covered by tests
https://codecov.io/gh/aiogram/aiogram/pull/284/diff?src=pr&el=tree#D1-44

aiogram/api/client/session/aiohttp.py
uwinx: library supports it too. I think, we shouldn't care about proxy. Aiohttp does not support socks proxies, AFAIK.

aiogram/api/client/session/base.py
uwinx: Yep, `config` would be nicer.
I guess, I didn't want to mess session's namespace and keep "configs" as simple namespace.

b0g3r: By the way, why is it named config and placed in BaseSession? It's only my opinion, but private attributes like `_connector_type` and `_connector_params` in AiohttpSession would be better

aiogram/api/client/session/aiohttp.py
```diff
@@ -1,29 +1,72 @@
...
-        super(AiohttpSession, self).__init__(api=api, json_loads=json_loads, json_dumps=json_dumps)
+        super(AiohttpSession, self).__init__(
+            api=api,
+            json_loads=json_loads,
+            json_dumps=json_dumps,
+            proxy=proxy
+        )
         self._session: Optional[ClientSession] = None
+        self.cfg.connector_type = TCPConnector
+        self.cfg.connector_init = cast(Dict[str, Any], {})
```
b0g3r: Please, use explicit types instead of casting:
```suggestion
        self.cfg.connector_init: Dict[str, Any] = {}
```

I'm not sure about compatibility with simple namespaces, but it avoids type errors in the future.

Also, check the possibility to add typing for connector_type, like `: Type[BaseConnector]` or `: Type[TCPConnection]`

aiogram/api/client/session/aiohttp.py
```diff
@@ -1,29 +1,72 @@
...
+                if not username:
+                    username = proxy_auth.login
+                if not password:
+                    password = proxy_auth.password
+
+            self.cfg.connector_init.update(
+                dict(
+                    proxy_type=proxy_type, host=host, port=port,
+                    username=username, password=password,
+                    rdns=True,
```
b0g3r: What is rdns? Should we give the ability to change this param by user?

aiogram/api/client/session/aiohttp.py
uwinx: Should I make different test passing URL as `proxy`?

aiogram/api/client/session/aiohttp.py
uwinx: 1. You cannot do that.
2. It infers type from `aiohttp.TCPConnector`, I'd consider annotating it to `aiohttp.BaseConnector`

aiogram/api/client/session/aiohttp.py
uwinx: `rdns` is `use no resolver` boolean. Current aiogram-dev-2.x implements it this way.

b0g3r: Thank you for your answer!

aiogram/api/client/session/aiohttp.py
b0g3r: Yes

aiogram/api/client/session/aiohttp.py
b0g3r: About youtrack issue: the comment says that attribute and type declaration should be in `__init__` or `__new__` methods, so it's our case

b0g3r: Casting is a problem because it can hide accidental errors, e.g.:
```
new_attr = cast(int, .1)  # no error
new_attr: int = .1  # error
```

aiogram/api/client/session/base.py
uwinx: It's inner configuration that designed to be in one namespace as I commented above.

aiogram/api/client/session/aiohttp.py
uwinx: > About youtrack issue: the comment says that attribute and type declaration should be in `__init__` or `__new__` methods, so it's our case

This is not our case. We are not in the `types.SimpleNamespace::__init__`

b0g3r: So, it's definitely better to not use `SimpleNamespace` as an attribute storage

b0g3r: It can be illustrated with example:
```python
from types import SimpleNamespace
from typing import cast

class AFewMethods:
    def __init__(self):
        self.namespace = SimpleNamespace()
        self.namespace.a: int = 1
        self.namespace.b = cast(int, 1)
        reveal_type(self.namespace.a) # note: Revealed type is 'builtins.int'
        reveal_type(self.namespace.b) # note: Revealed type is 'builtins.int'

    def another_method(self):
        reveal_type(self.namespace.a) # note: Revealed type is 'Any'
        reveal_type(self.namespace.b) # note: Revealed type is 'Any'
        pass
```

uwinx: The main idea behind using `SimpleNamespace` is not messing `BaseSession`'s namespace. Treat it as `Dict[str, Any]`. I found `SimpeNamespace` "nicer", because it's a facade on top of `dict`

[documentation on topic](https://docs.python.org/3/library/types.html#types.SimpleNamespace)

aiogram/api/client/session/base.py
uwinx: :disappointed: Since the `SimpleNamespace` was not met as I wanted, I'll just set new attributes starting with underscore. Thanks for the opinions and arguments on implementation disadvantages.

aiogram/api/client/session/aiohttp.py
b0g3r: As I promised, here are a few options on how to achieve typing without private attributes for each param:
```python
from typing import (
    NamedTuple,
    Optional,
    TypedDict,
)


class ConfigTypedDict(TypedDict, total=False):
    param: int
    another: str


class ExampleTypedDict:
    def __init__(self):
        self.config: ConfigTypedDict = {}
        self.config['param'] = 1


class ConfigNamedTuple(NamedTuple):
    param: int
    another: str

class ExampleNamedTuple:
    def __init__(self):
        # immutable, so should be replaced every time when we want a change config
        self.config = ConfigNamedTuple(param=1, another='oh')


class ConfigJustClass:
    def __init__(self, param: int, another: str):
        self.param = param
        self.another = another

class ExampleJustClass:
    def __init__(self):
        self.config = ConfigJustClass(param=1, another='oh')
        self.config.param = 2
```

PR #319, opened 2020-05-03
Fix behavior when a single string is passed as ID

Senpos:
# Description

If you pass a single string as ID (either user_id or chat_id) to IDFilter/AdminFilter - it will be treated as Iterable, as a result:
- if it is a group chat id (start with `-`) - error will be raised, since `-` can't be converted to `int`
- if it is a user_id or a personal chat (positive integer) - ID filter will not work, since string will be splitted into characters

`self.user_id` and `self.chat_id` are now `Set[int]`, since there's no point to allow passing non-unique IDs and order does not really matter.

**Examples:**
`IDFilter('-1234')` -> `ValueError: invalid literal for int() with base 10: '-'`
`IDFilter('1234')` -> filter will search for users/chats with ids `[1, 2, 3, 4]`

## Type of change

# How Has This Been Tested?
```python
@dp.message_handler(IDFilter('-64856280'))
@dp.message_handler(IDFilter('64856280'))
@dp.message_handler(IDFilter(-64856280))
@dp.message_handler(IDFilter(64856280))
@dp.message_handler(IDFilter(['-64856280']))
@dp.message_handler(IDFilter([-64856280]))
@dp.message_handler(IDFilter(['-64856280', '-64856280']))
@dp.message_handler(IDFilter(['-64856280', -64856280]))
@dp.message_handler(IDFilter([-64856280, -64856280]))
@dp.message_handler(IDFilter(iter(['-64856280'])))
async def id_filter_demo_handler(message: types.Message):
    await message.answer(f'User: {message.from_user.id}, Chat: {message.chat.id}')
```
All of the filters above work (use one at a time)
Added tests for cases I mentioned above

**Test Configuration**:
* Operating System: Windows 7 x64
* Python version: 3.7.5

# Checklist:
- [x] I have made corresponding changes to the documentation (it will be autogenerated)

### DIFF
--- a/aiogram/dispatcher/filters/builtin.py
+++ b/aiogram/dispatcher/filters/builtin.py
@@ -15,14 +15,14 @@ from aiogram.types import CallbackQuery, Message, InlineQuery, Poll, ChatType
 ChatIDArgumentType = typing.Union[typing.Iterable[typing.Union[int, str]], str, int]
 
 
-def extract_chat_ids(id_filter_argument: ChatIDArgumentType) -> typing.Set[int]:
+def extract_chat_ids(chat_id: ChatIDArgumentType) -> typing.Set[int]:
     # since "str" is also an "Iterable", we have to check for it first
-    if isinstance(id_filter_argument, str):
-        return {int(id_filter_argument),}
-    if isinstance(id_filter_argument, Iterable):
-        return {int(item) for (item) in id_filter_argument}
+    if isinstance(chat_id, str):
+        return {int(chat_id), }
+    if isinstance(chat_id, Iterable):
+        return {int(item) for (item) in chat_id}
     # the last possible type is a single "int"
-    return {id_filter_argument,}
+    return {chat_id, }
 
 
 class Command(Filter):
@@ -622,22 +622,20 @@ class AdminFilter(Filter):
     is_chat_admin is required for InlineQuery.
     """
 
-    def __init__(self, is_chat_admin: Optional[Union[Iterable[Union[int, str]], str, int, bool]] = None):
+    def __init__(self, is_chat_admin: Optional[Union[ChatIDArgumentType, bool]] = None):
         self._check_current = False
         self._chat_ids = None
 
         if is_chat_admin is False:
             raise ValueError("is_chat_admin cannot be False")
 
-        if is_chat_admin:
-            if isinstance(is_chat_admin, bool):
-                self._check_current = is_chat_admin
-            if isinstance(is_chat_admin, Iterable):
-                self._chat_ids = list(is_chat_admin)
-            else:
-                self._chat_ids = [is_chat_admin]
-        else:
+        if not is_chat_admin:
             self._check_current = True
+            return
+
+        if isinstance(is_chat_admin, bool):
+            self._check_current = is_chat_admin
+        self._chat_ids = extract_chat_ids(is_chat_admin)
 
     @classmethod
     def validate(cls, full_config: typing.Dict[str, typing.Any]) -> typing.Optional[typing.Dict[str, typing.Any]]:
--- a/aiogram/dispatcher/filters/builtin.py
+++ b/aiogram/dispatcher/filters/builtin.py
@@ -556,8 +556,8 @@ class ExceptionsFilter(BoundFilter):
         except:
             return False
 
-class IDFilter(Filter):
 
+class IDFilter(Filter):
     def __init__(self,
                  user_id: Optional[ChatIDArgumentType] = None,
                  chat_id: Optional[ChatIDArgumentType] = None,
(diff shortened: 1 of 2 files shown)


aiogram/dispatcher/filters/builtin.py
```diff
@@ -12,6 +12,19 @@
 from aiogram.types import CallbackQuery, Message, InlineQuery, Poll, ChatType
 
 
+IDFilterArgumentType = typing.Union[typing.Iterable[typing.Union[int, str]], str, int]
+
+
+def extract_filter_ids(id_filter_argument: IDFilterArgumentType) -> typing.Set[int]:
```
evgfilim1, line 18: I think it would be better to put this method into `IDFilter` as `@staticmethod`

Senpos: Should it be unique to IDFilter though?
You may use the same type & helper function in other filters.

This can be used in AdminFilter with a small changes.

evgfilim1: Hmm.. yes, it can.

evgfilim1:
Looks like `AdminFilter` has the same bug: https://github.com/aiogram/aiogram/blob/ac7758aeb03bf186a701fde0dd2bce4254165d82/aiogram/dispatcher/filters/builtin.py#L635
It would be nice to fix it too :)

Birdi7:
> # How Has This Been Tested?
> ```python
> @dp.message_handler(IDFilter('-64856280'))
> @dp.message_handler(IDFilter('64856280'))
> @dp.message_handler(IDFilter(-64856280))
> @dp.message_handler(IDFilter(64856280))
> @dp.message_handler(IDFilter(['-64856280']))
> @dp.message_handler(IDFilter([-64856280]))
> @dp.message_handler(IDFilter(['-64856280', '-64856280']))
> @dp.message_handler(IDFilter(['-64856280', -64856280]))
> @dp.message_handler(IDFilter([-64856280, -64856280]))
> @dp.message_handler(IDFilter(iter(['-64856280']))
> ```

What do you think about converting this to a unit test?

aiogram/dispatcher/filters/builtin.py
Birdi7: also, for me, the naming is a bit obscure. `extract_filter_ids` says about `filter_ids` like about a special type of `ids`. As these are just regular chat ids, I would think about changing the name to `extract_chat_ids`, but that's up to you

aiogram/dispatcher/filters/builtin.py
```diff
@@ -12,6 +12,19 @@
 from aiogram.types import CallbackQuery, Message, InlineQuery, Poll, ChatType
 
 
+IDFilterArgumentType = typing.Union[typing.Iterable[typing.Union[int, str]], str, int]
```
uwinx, line 15: what about:
```
_UserID = typing.Union[str, int]
IDFilterArgumentType = typing.Union[typing.Iterable[_UserID], _UserID]
```
?

Birdi7: I doubt if this good name. `IDFilterArgumentType` implies that this type is used only in IDFilter, but that's not the case here, is it? But LGTM
```python3
_ID = TypeVar("_ID", int, str)
IDArgumentType = Union[Sequence[_ID], _ID]
```

uwinx: I don't think name matters here, type is internal. However I agree sounds nicer if shorter.

evgfilim1:
@Senpos any progress?

aiogram/dispatcher/filters/builtin.py
evgfilim1: > shouldn't we consider moving it to `utils`

@Birdi7 I think no as it's internal helper for this file (for now).

aiogram/dispatcher/filters/builtin.py
Senpos: Changed type name to `ChatIDArgumentType`

aiogram/dispatcher/filters/builtin.py
Birdi7: this function also can be useful for users of lib in some cases, but ok

aiogram/dispatcher/filters/builtin.py
```diff
@@ -12,6 +12,19 @@
 from aiogram.types import CallbackQuery, Message, InlineQuery, Poll, ChatType
 
 
+ChatIDArgumentType = typing.Union[typing.Iterable[typing.Union[int, str]], str, int]
```
uwinx, line 15: just a tiny point :)
we've already imported values for (Union, Iterable) types below (line 6)

PR #327, opened 2020-05-09
AIOG-T-25 refactor(handler): rename observers

uwinx:
Rename observers but with backward compatibility, relevant documentation

## Type of change

- [x] Documentation
- [x] Breaking change

# Checklist:

## Implementation

The reason I didn't use `aiogram.utils.deprecated::deprecated` is I didn't want to pollute property's lower decorator and make codebase "nasty" :-)
Deprecated properties are generated with pattern. Documentation/tests was checked by me. New test to check if function is deprecated or removed in older/newer versions was added too.

### DIFF
--- a/tests/deprecated.py
+++ b/tests/deprecated.py
@@ -1,0 +1,26 @@
+from contextlib import contextmanager
+from typing import Type
+
+import pytest
+from packaging import version
+
+import aiogram
+
+
+@contextmanager
+def check_deprecated(
+    max_version: str, exception: Type[Exception], warning: Type[Warning] = DeprecationWarning,
+) -> None:
+    """
+    Should be used for modules that are being deprecated or already removed from aiogram
+    """
+
+    parsed_max_version = version.parse(max_version)
+    current_version = version.parse(aiogram.__version__)
+
+    if parsed_max_version <= current_version:
+        with pytest.raises(exception):
+            yield
+    else:
+        with pytest.warns(warning, match=max_version):
+            yield
--- a/tests/test_dispatcher/test_deprecated.py
+++ b/tests/test_dispatcher/test_deprecated.py
@@ -23,17 +23,14 @@ OBSERVERS = {
 def test_deprecated_handlers_name():
     from aiogram import __version__
 
-    minor_partial = int(__version__.split(".")[1])
+    major, minor = map(int, __version__.split(".")[:-1])
 
-    if minor_partial >= 2:
+    if minor >= 2 and major >= 3:  # version >=3.2.*
         do_assert = pytest.raises(AttributeError)
-    else:
+    else:  # for versions <=3.2.* (we don't care if major is lesser than `3`)
         do_assert = pytest.warns(DeprecationWarning)
 
     router = Router()
-
-    async def _(__):
-        ...
 
     with do_assert:
         for decor in OBSERVERS:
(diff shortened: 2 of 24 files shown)


tests/test_dispatcher/test_deprecated.py
```diff
@@ -0,0 +1,46 @@
...
+
+    if minor >= 2 and major >= 3:  # version >=3.2.*
+        do_assert = pytest.raises(AttributeError)
+    else:  # for versions <=3.2.* (we don't care if major is lesser than `3`)
+        do_assert = pytest.warns(DeprecationWarning)
+
+    router = Router()
+
+    with do_assert:
+        for decor in OBSERVERS:
```
JrooTJunior, line 36: Replace this for-loop with [parametrized](https://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions) test please

tests/test_dispatcher/test_deprecated.py
```diff
@@ -0,0 +1,46 @@
...
+    "update",
+}
+
+
+def test_deprecated_handlers_name():
+    from aiogram import __version__
+
+    major, minor = map(int, __version__.split(".")[:-1])
+
+    if minor >= 2 and major >= 3:  # version >=3.2.*
```
JrooTJunior, line 28: I think this validation can be moved to separated function because it can be used in other places

Make something like this:
```python
@contextmanager
def check_deprecated(
    max_version: str,
    warning: Type[Warning] = DeprecationWarning,
    exception: Type[Exception] = AttributeError,
) -> None:
    parsed_max_version = version.parse(max_version)
    current_version = version.parse(aiogram.__version__)

    if parsed_max_version < current_version:
        with pytest.raises(exception):
            yield
    else:
        with pytest.warns(warning, match=max_version):
            yield
```

In this case `packaging` should be added as development dependency (`poetry add -D packaging`).

And then it can be used like this:
```python
with check_deprecated('3.2'):
    getattr(<obj>, <attribute>)
```

tests/test_dispatcher/test_deprecated.py
```diff
@@ -0,0 +1,46 @@
...
+    else:  # for versions <=3.2.* (we don't care if major is lesser than `3`)
+        do_assert = pytest.warns(DeprecationWarning)
+
+    router = Router()
+
+    with do_assert:
+        for decor in OBSERVERS:
+            getattr(router, decor + "_handler")
+
+        assert all(
```
JrooTJunior, line 39: When you make this test parametrized you can use something like this:

```python3
with check_deprecated('3.2'):
    observer = getattr(router, observer_name)
    assert isinstance(observer, TelegramEventObserver)
```

And i think other validations is no more needed

tests/deprecated.py
```diff
@@ -0,0 +1,26 @@
...
+
+import pytest
+from packaging import version
+
+import aiogram
+
+
+@contextmanager
+def check_deprecated(
+    max_version: str, exception: Type[Exception], warning: Type[Warning] = DeprecationWarning,
```
JrooTJunior, line 12: Is only question: Why exception here without default value?

uwinx: `check_deprecated` is generic function for different features, so I thought passing required exception type makes sense in different cases.

JrooTJunior: tnx

PR #332, opened 2020-05-16
Download feature and URLInputFile

gabbhack:
# Description

- Rename BaseBot to Bot in tests
- Add Downloadable type (it's actually Protocol)
- Add download_file and download method
- Add docs about download
- Add URLInputFile

This PR is a continuation of #282.
After discussions, it was decided to abandon `Downloadable` as mixin in favor of the `bot.download()` method.

## Type of change

# How Has This Been Tested?

- [x] Add tests for download_file
- [x] Add tests for download
- [x] Add tests for URLInputFile

**Test Configuration**:
* Operating System: Kubuntu 20.04
* Python version: 3.8.2

# Checklist:


gabbhack:
Need to work with Related pages in docs

aiogram/api/client/bot.py
```diff
@@ -167,6 +181,93 @@ async def close(self) -> None:
...
+
+        :param file: file_id or Downloadable object
+        :param destination: Filename, file path or instance of :class:`io.IOBase`. For e.g. :class:`io.BytesIO`, defaults to None
+        :param timeout: Total timeout in seconds, defaults to 30
+        :param chunk_size: File chunks size, defaults to 64 kb
+        :param seek: Go to start of file when downloading is finished. Used only for destination with :class:`typing.BinaryIO` type, defaults to True
+        """
+        if isinstance(file, str):
+            file_id = file
+        elif hasattr(file, "file_id"):
```
uwinx, line 258: seems like a redundant check. `getattr` can be used :D

aiogram/api/client/bot.py
```diff
@@ -167,6 +181,93 @@ async def close(self) -> None:
...
+            file_id = file.file_id
+        else:
+            raise TypeError("file can only be of the string or Downloadable type")
+
+        _file = await self.get_file(file_id)
+        # https://github.com/aiogram/aiogram/pull/282/files#r394110017
+        file_path = cast(str, _file.file_path)
```
uwinx, line 265: what about `# type: str`?

aiogram/api/client/bot.py
```diff
@@ -167,6 +181,93 @@ async def close(self) -> None:
...
+    async def download_file(
+        self,
+        file_path: str,
+        destination: Optional[Union[BinaryIO, pathlib.Path, str]] = None,
+        timeout: int = 30,
+        chunk_size: int = 65536,
+        seek: bool = True,
+    ) -> Optional[BinaryIO]:
+        """
+        Download file by file_path to destination.
```
uwinx, line 212: nice explanation, can we have it in one place only? :)

aiogram/api/client/bot.py
```diff
@@ -167,6 +181,93 @@ async def close(self) -> None:
         """
         await self.session.close()
 
+    @staticmethod
```
uwinx, line 184: can we move `staticmethod`s outside of class if they're used internally only? we don't inherit from `Bot` if it's name mangling, do we? :thinking:

docs/api/downloading_files.md
```diff
@@ -0,0 +1,104 @@
...
+await bot.download_file(file_path, "text.txt")
+```
+
+### Download file to binary I/O object
+
+To download file to binary I/O object, you must specify an object with the `#!python3 typing.BinaryIO` type or use the default (`#!python3 None`) value.
+
+In the first case, the function will return your object:
+```python3
+my_object = MyBinaryIO()
```
uwinx, line 49: let's use something real for example :)

aiogram/api/client/bot.py
gabbhack: Oh, I don't get it. Can you give an example?

aiogram/api/client/bot.py
gabbhack: mypy seems to be ignoring this comment

aiogram/api/client/bot.py
gabbhack: do you mean to add that file_path can be obtained from the getFile method?

docs/api/downloading_files.md
gabbhack: I can't think of anything other than to use `io.BytesIO()` instead of `MyBinaryIO()`, but why would the user need it if `io.BytesIO` is returned by default.
To be honest, I haven't had to use my own IO object, so I'm waiting for an example from you)

aiogram/api/client/bot.py
evgfilim1: @gabbhack 
```python
if isinstance(file, str):
    ...
else:
    file_id = getattr(file, "file_id", None)
    if file_id is None:
        raise TypeError(...)
```

aiogram/api/client/bot.py
uwinx: Sorry, I was writing about docstring. can we have it in `docs/*` only? not to edit many times.

gabbhack: But other methods are described in both docstring and docs/*
I don't think that "syncing" documentation takes much time, and that this "syncing" will be performed frequently. And it is more convenient for the user not to open the documentation on the site every time.

uwinx: I understand, but sometimes we understand to do such things :D
aren't other methods generated?

Yep, this is not very necessary, I get it

PR #334, opened 2020-05-17
Added always_private argument to url() method

MasterGroosha:
# Description

There are some cases when you want to be sure that your link will work regardless of chat username. This change makes it so you can force creation of private link

## Type of change

# How Has This Been Tested?

Not tested

# Checklist:

Lol no, I just edited on github

### DIFF
--- a/aiogram/types/message.py
+++ b/aiogram/types/message.py
@@ -228,22 +228,23 @@ class Message(base.TelegramObject):
         return self.parse_entities()
 
     @property
-    def url(self) -> str:
+    def url(self, always_private=False) -> str:
         """
         Get URL for the message
 
+        :param always_private: Force generate a private URL even when username is available
         :return: str
         """
         if ChatType.is_private(self.chat):
             raise TypeError('Invalid chat type!')
 
         url = 'https://t.me/'
-        if self.chat.username:
-            # Generates public link
-            url += f'{self.chat.username}/'
-        else:
+        if not self.chat.username or always_private:
             # Generates private link available for chat members
             url += f'c/{self.chat.shifted_id}/'
+        else:
+            # Generates public link
+            url += f'{self.chat.username}/'
         url += f'{self.message_id}'
 
         return url


MasterGroosha:
Instead of "always_private" it can be named as "force_private". Maybe it will fit better

aiogram/types/message.py
```diff
@@ -228,22 +228,23 @@ def html_text(self) -> str:
         return self.parse_entities()
 
     @property
-    def url(self) -> str:
+    def url(self, always_private=False) -> str:
```
uwinx, line 231: @MasterGroosha, `url` is a `property`, so having any parameter except `self` in its getter is impossible. descriptor's getter has to have the following signature `(self, instance, owner) -> T` :)

You may take a look at https://docs.python.org/3/howto/descriptor.html and remember descriptors.

evgfilim1: Oh, haven't noticed. :eyes:

MasterGroosha:
@uwinx aiogram code is too overkill for me in terms of number of places you need to edit to add a damn single method, feel free to do that yourself.

evgfilim1:
> number of places you need to edit to add a damn single method

@MasterGroosha For this particular feature you can just add a new property (as suggested by @uwinx) and slightly modify existing one to avoid code duplication. Everything can be done in the only one file. :shrug:

uwinx:
@MasterGroosha, I'm very disappointed to see you given up and this PR closed. We appreciate your help. However, we can't approve changes that simply won't work. In our case I have suggested a solution, which would have brought exactly one new property and described why wouldn't it work as we want.

Thanks, reopen or comment on anytime as you wish.

MasterGroosha:
Unfortunately, in my opinion this doesn’t look like a good solution. A better solution would be removing `url` property and adding a new method, for example, `get_url()`, which will be used as a unified way to get message URL.

However, this would be not backwards compatible as it will break the existing code. Also it would require writing tests which I’m not good with.

TL;DR: just adding a new property will lead to code duplication, potentially harming the framework in long run.

PR #336, opened 2020-05-25
feat(helpers): implement new descriptor with default value getter

uwinx:
# Description

Implement descriptor with pre-set default value. Similar to `property` descriptor, except narrowed to our use. At this point internal feature for in-library use.

This feature is not actually necessary. We may discuss if usage is obvious and should stay in.

## Type of change

# Checklist:


aiogram/api/client/session/base.py
```diff
@@ -20,47 +21,12 @@ class BaseSession(abc.ABC):
...
-
-    @timeout.deleter
-    def timeout(self) -> None:
-        if hasattr(self, "_timeout"):
-            del self._timeout
+    api: DefaultProperty[TelegramAPIServer] = DefaultProperty(PRODUCTION)
+    json_loads: DefaultProperty[_JsonLoads] = DefaultProperty(json.loads)
+    json_dumps: DefaultProperty[_JsonDumps] = DefaultProperty(json.dumps)
+    timeout: DefaultProperty[float] = DefaultProperty(
+        fget=lambda self: float(self.__class__.default_timeout)
```
JrooTJunior, line 28: What does mean `fget`?
This question is only about naming.

aiogram/utils/helper.py
```diff
@@ -233,3 +235,60 @@ def all(cls) -> List[str]:
...
+
+        return cast(T, getattr(instance, self.name, self.fget(instance)))  # type: ignore
+
+    def __set__(self, instance: Any, value: T) -> None:
+        if instance is None or isinstance(instance, type):
+            raise AttributeError(
+                "Instance cannot be class or None. Setter must be called from a class."
+            )
+        self._raise_if_name_never_set(instance)
+        setattr(instance, self.name, value)  # type: ignore
```
JrooTJunior, line 286: weakref can be used here instead of extending instance namespace? (Limitations for weakref - instance should be hashable)

aiogram/utils/helper.py
```diff
@@ -233,3 +235,60 @@ def all(cls) -> List[str]:
             else:
                 result.append(value)
         return result
+
+
+class DefaultProperty(Generic[T]):
```
JrooTJunior, line 240: What do you think about renaming to `Default`?

aiogram/utils/helper.py
```diff
@@ -233,3 +235,60 @@ def all(cls) -> List[str]:
...
+        "name",
+        "fget",
+    )
+
+    def __init__(
+        self,
+        default: Optional[T] = None,
+        *,
+        name_resolver: Callable[[str], str] = lambda s: "_" + s,
+        fget: Optional[Callable[[Any], T]] = None,
```
JrooTJunior, line 256: `fget` means something like "factory get"?
So if answer for my question - "Yes", this argument should be renamed to `factory`

uwinx: it's actually came from python's built-in `property` parameter `fget`. So no, it's a function to be used for getting an attribute value

aiogram/utils/helper.py
uwinx: It's a property-like descriptor and intended for in-class only use, so I think `Property` suffix clarifies where does `DefaultProperty` belong.

On the other hand, we declare it as an internal feature -- sure we can rename it + adding nice docstring :)

aiogram/utils/helper.py
JrooTJunior: ok, you can leave it as it is

aiogram/utils/helper.py
JrooTJunior: So... As weakref take too much limitations where this descriptor can be used don't need to use it :man_shrugging:

PR #339, opened 2020-05-27
feat(handler): requirement class

uwinx:
# Description

First and not very smart solution for handler requirements.

What is requirements in aiogram handler? They're inspired by awesome [fastapi](https://github.com/tiangolo/fastapi)'s [Dependency](https://fastapi.tiangolo.com/tutorial/dependencies/). However, in aiogram case, I don't want "Requirements" to be "dependencies", when they're really just a sweet addition. Aiogram will take care of "initializing" those requirements by some set of rules.  

## Type of change

- [x] New feature (breaking change which adds functionality)

# How Has This Been Tested?

You can take a look at very simple [usage](https://pastebin.com/bX8SUG9h) and play out.

**Test Configuration**:
* Python version: 3.8

# Checklist:

### Details that can be unclear

We create cache, exit_stack for every event processing operation. Cache and stack are intermediate internal keys used to solve requirements (they're probably exposed in new middlewares). Filters can have requirements too. generators and asynchronous generators are nicely solved with `contextmanager` which, I guess, makes their usage "clear".

### Limitations

Currently requirement's callback cannot have any parameter which is not `Requirement`


aiogram/dispatcher/event/handler.py
```diff
@@ -1,45 +1,93 @@
...
+    return isinstance(handler, type) and issubclass(handler, BaseHandler)
+
 
 @dataclass
 class CallableMixin:
     callback: HandlerType
     awaitable: bool = field(init=False)
     spec: inspect.FullArgSpec = field(init=False)
 
+    __reqs__: Dict[str, Requirement[Any]] = field(init=False)
```
JrooTJunior, line 38: Why this attribute named like magic?

aiogram/dispatcher/event/handler.py
```diff
@@ -1,45 +1,93 @@
...
+
     def __post_init__(self) -> None:
         callback = inspect.unwrap(self.callback)
         self.awaitable = inspect.isawaitable(callback) or inspect.iscoroutinefunction(callback)
         if isinstance(callback, BaseFilter):
-            # Pydantic 1.5 has incorrect signature generator
-            # Issue: https://github.com/samuelcolvin/pydantic/issues/1419
-            # Fixes: https://github.com/samuelcolvin/pydantic/pull/1427
-            # TODO: Remove this temporary fix
```
JrooTJunior, line 31: This TODO should not be removed because bug fixes is not yet released.

aiogram/dispatcher/event/handler.py
```diff
@@ -1,45 +1,93 @@
...
-            callback = inspect.unwrap(callback.__call__)
+            callback = callback.__call__
```
JrooTJunior, line 44: Here `inspect.unwrap` should be used because callback can be wrapped by decorators and can modify method specs and removing `unwrap` can generate unexpected behavior for class based filters with decorated `__call__` method

aiogram/dispatcher/event/handler.py
```diff
@@ -59,11 +107,6 @@ class HandlerObject(CallableMixin):
     callback: HandlerType
     filters: Optional[List[FilterObject]] = None
 
-    def __post_init__(self) -> None:
-        super(HandlerObject, self).__post_init__()
-        if inspect.isclass(self.callback) and issubclass(self.callback, BaseHandler):  # type: ignore
-            self.awaitable = True
```
JrooTJunior, line 65: How do you plans to detect awaitable callbacks?

aiogram/dispatcher/requirement.py
```diff
@@ -0,0 +1,173 @@
...
+        yield context_manager.__enter__()
+    except Exception as e:
+        err = context_manager.__exit__(type(e), e, None)
+        if not err:
+            raise e
+    else:
+        context_manager.__exit__(None, None, None)
+
+
+class Requirement(abc.ABC, Generic[T]):
```
JrooTJunior, line 45: That is `BaseRequirement` or `AbstractRequirement`

aiogram/dispatcher/requirement.py
```diff
@@ -0,0 +1,173 @@
...
+        if self.use_cache and self.cache_key in cache_dict:
+            return cast(T, cache_dict[self.cache_key])
+        else:
+            result = await initialize_callable_requirement(self, data, stack)
+            if self.use_cache:
+                cache_dict[self.cache_key] = result
+            return result
+
+
+class CachedRequirement(Requirement[T]):
```
JrooTJunior, line 119: Cache mode can be controlled via init arguments without adding one more class?
Users can be confused for too much names of single feature

For example by `use_cache: bool = True`

or by `cache_key: Union[bool, CacheKeyType] = True`
in this case when value is True - generate key automatically, when False - don't use caching.

aiogram/dispatcher/requirement.py
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+
+def require(
+    what: _RequiredCallback[T],
+    *,
+    cache_key: Optional[CacheKeyType] = None,
+    use_cache: bool = True,
+) -> T:
+    return CallableRequirement(what, cache_key=cache_key, use_cache=use_cache)  # type: ignore
```
JrooTJunior, line 173: Why this factory is needed? Why `Requirement(...)` can't be used directly?

pyproject.toml
```diff
@@ -34,7 +34,7 @@ classifiers = [
 [tool.poetry.dependencies]
 python = "^3.7"
 aiohttp = "^3.6"
-pydantic = "^1.5"
+pydantic = "^1.5.1"
```
JrooTJunior, line 37: Still waiting for 1.5.2 or 1.6... hrr
Not all bugfixes is released with 1.5.1

aiogram/dispatcher/event/handler.py
uwinx: Honestly, I don't recall why if I named it as a dunder attribute. Noted.

aiogram/dispatcher/event/handler.py
uwinx: pydantic1.5.1 shipped fix for this, didn't they?

aiogram/dispatcher/event/handler.py
uwinx: callback are naturally inspected in `CallableMixin`. I didn't actually change the behaviour of inspection, did I? :thinking:
This is check determined whether `BaseHandler` subclass was passed to `HandlerObject` and seemed redundant to me, since we can check callback once in `CallableMixin`'s `__post_init__`

pyproject.toml
uwinx: Got it. Gotta rollback our pydantic-related related changes.
Tests are passing with 1.5.1 tho.

aiogram/dispatcher/requirement.py
uwinx: I wanted to differentiate requirements. I clearly need to rethink that now.

aiogram/dispatcher/requirement.py
uwinx: You're right. However, keep in mind that we have bunch of inspections of `callable` in `CallableRequirement.__init__`, whereas `CachedRequirement` is just a cache-key holder.

I will definitely,  rethink design for requirements

aiogram/dispatcher/event/handler.py
```diff
@@ -1,45 +1,93 @@
...
+        else:
+            self.__reqs__ = get_reqs_from_callable(callable_=callback)
+
     def _prepare_kwargs(self, kwargs: Dict[str, Any]) -> Dict[str, Any]:
         if self.spec.varkw:
             return kwargs
 
         return {k: v for k, v in kwargs.items() if k in self.spec.args}
 
     async def call(self, *args: Any, **kwargs: Any) -> Any:
```
uwinx, line 61: @JrooTJunior, shall we change the signature `(*args: Any, **kwargs: Any)` to `(event: Any, **data: Any)`

aiogram/dispatcher/event/handler.py
JrooTJunior: Nope. Related with https://github.com/samuelcolvin/pydantic/pull/1466 and is not in release

aiogram/dispatcher/event/handler.py
JrooTJunior: Yep

aiogram/dispatcher/event/handler.py
JrooTJunior: I need to check it

aiogram/dispatcher/event/handler.py
uwinx: Uh, strange then. Tests are running well with 1.5.1 without `inspect.unwrap` and failing with it.

pyproject.toml
uwinx: resolved.

aiogram/dispatcher/event/handler.py
uwinx: seems like can be resolved as of pydantic 1.6.x

uwinx:
Closing this after a lot of thinking. This feature unnecessarily makes it hard to maintain. If someone wants their function called before handler gets executed, it should not be done on a new level of update handling from aiogram. It can be done in middlewares/filters already.

PR #340, opened 2020-05-27
Added fields to ChatMember

lyteloli:
# Description

## Type of change

# How Has This Been Tested?

- [x] Pytest of is_chat_admin()
- [x] Pytest of is_chat_member()

**Test Configuration**:
* Operating System: MacOS Catalina 10.15.4
* Python version: 3.8

# Checklist:

### DIFF
--- a/aiogram/api/types/chat_member.py
+++ b/aiogram/api/types/chat_member.py
@@ -4,6 +4,7 @@ import datetime
 from typing import TYPE_CHECKING, Optional, Union
 
 from .base import TelegramObject
+from ...utils import helper
 
 if TYPE_CHECKING:  # pragma: no cover
     from .user import User
@@ -65,3 +66,31 @@ class ChatMember(TelegramObject):
     inline bots"""
     can_add_web_page_previews: Optional[bool] = None
     """Restricted only. True, if the user is allowed to add web page previews to their messages"""
+
+    def is_chat_admin(self) -> bool:
+        return ChatMemberStatus.is_chat_admin(self.status)
+
+    def is_chat_member(self) -> bool:
+        return ChatMemberStatus.is_chat_member(self.status)
+
+
+class ChatMemberStatus(helper.Helper):
+    """
+    Chat member status
+    """
+    mode = helper.HelperMode.lowercase
+
+    CREATOR = helper.Item()  # creator
+    ADMINISTRATOR = helper.Item()  # administrator
+    MEMBER = helper.Item()  # member
+    RESTRICTED = helper.Item()  # restricted
+    LEFT = helper.Item()  # left
+    KICKED = helper.Item()  # kicked
+
+    @classmethod
+    def is_chat_admin(cls, role: str) -> bool:
+        return role in [cls.ADMINISTRATOR, cls.CREATOR]
+
+    @classmethod
+    def is_chat_member(cls, role: str) -> bool:
+        return role in [cls.MEMBER, cls.ADMINISTRATOR, cls.CREATOR, cls.RESTRICTED]
--- a/aiogram/api/types/chat_member.py
+++ b/aiogram/api/types/chat_member.py
@@ -67,9 +67,11 @@ class ChatMember(TelegramObject):
     can_add_web_page_previews: Optional[bool] = None
     """Restricted only. True, if the user is allowed to add web page previews to their messages"""
 
+    @property
     def is_chat_admin(self) -> bool:
         return self.status in {ChatMemberStatus.CREATOR, ChatMemberStatus.ADMINISTRATOR}
 
+    @property
     def is_chat_member(self) -> bool:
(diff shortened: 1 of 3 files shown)


docs/api/types/chat_member.md
```diff
@@ -23,6 +23,7 @@ This object contains information about one member of a chat.
 | `can_invite_users` | `#!python Optional[bool]` | Optional. Administrators and restricted only. True, if the user is allowed to invite new users to the chat |
 | `can_pin_messages` | `#!python Optional[bool]` | Optional. Administrators and restricted only. True, if the user is allowed to pin messages; groups and supergroups only |
 | `is_member` | `#!python Optional[bool]` | Optional. Restricted only. True, if the user is a member of the chat at the moment of the request |
+| `is_chat_admin` | `#!python bool` | True if the user is administrator or creator of the chat |
```
JrooTJunior, line 26: Aliases/additional methods/properties should not be mixed with original attributes in types because in this listed attributes from original Bot API documentation. Also this table can be overwrited by code-gen.

Just add after "Attributes" table new section named "Extensions" and then add the table with new methods and helper class.
Helper can be described in section "ChatMemberStatus Helper"

aiogram/api/types/chat_member.py
```diff
@@ -65,3 +66,31 @@ class ChatMember(TelegramObject):
...
+    CREATOR = helper.Item()  # creator
+    ADMINISTRATOR = helper.Item()  # administrator
+    MEMBER = helper.Item()  # member
+    RESTRICTED = helper.Item()  # restricted
+    LEFT = helper.Item()  # left
+    KICKED = helper.Item()  # kicked
+
+    @classmethod
+    def is_chat_admin(cls, role: str) -> bool:
+        return role in [cls.ADMINISTRATOR, cls.CREATOR]
```
JrooTJunior, line 92:
```suggestion
        return role in {cls.ADMINISTRATOR, cls.CREATOR}
```

aiogram/api/types/chat_member.py
```diff
@@ -65,3 +66,31 @@ class ChatMember(TelegramObject):
...
+
+    @classmethod
+    def is_chat_member(cls, role: str) -> bool:
+        return role in [cls.MEMBER, cls.ADMINISTRATOR, cls.CREATOR, cls.RESTRICTED]
```
JrooTJunior, line 96:
```suggestion
        return role in {cls.MEMBER, cls.ADMINISTRATOR, cls.CREATOR, cls.RESTRICTED}
```

JrooTJunior:
And also what do you think about moving (logic) methods from helper class to original methods in ChatMember methods? (Single responsibility)

JrooTJunior:
I spoke only about moving methods from helper but not for removing helper :smile:

lyteloli:
> I spoke only about moving methods from helper but not for removing helper :smile: 

What would we need it for if it provides no available methods?

aiogram/api/types/chat_member.py
```diff
@@ -65,3 +65,9 @@ class ChatMember(TelegramObject):
     inline bots"""
     can_add_web_page_previews: Optional[bool] = None
     """Restricted only. True, if the user is allowed to add web page previews to their messages"""
+
+    def is_chat_admin(self) -> bool:
+        return self.status in {"creator", "administrator"}
+
+    def is_chat_member(self) -> bool:
+        return self.status not in {"left", "kicked"}
```
uwinx, line 73: how does it differ from `is_member` field?

aiogram/api/types/chat_member.py
```diff
@@ -65,3 +65,9 @@ class ChatMember(TelegramObject):
     inline bots"""
     can_add_web_page_previews: Optional[bool] = None
     """Restricted only. True, if the user is allowed to add web page previews to their messages"""
+
+    def is_chat_admin(self) -> bool:
+        return self.status in {"creator", "administrator"}
```
uwinx, line 70: what if we make it `property`?
by the way, we need to ensure `self.status` has the right case (I think, using `str.lower` would be enough. However there's `str.casefold` for both operands)

aiogram/api/types/chat_member.py
lyteloli: is_member is optional for restricted users only and returned by the api. True, if the user is a member of the chat at the moment of the request unlike is_chat_member which is evaluated in place.

lyteloli:
@uwinx I made it a property and personally I also thought to do so but in the TODO list it was a method so we shall wait for @JrooTJunior to review it again.

aiogram/api/types/chat_member.py
igorsereda: I think, `status` values should be enumerated.

JrooTJunior: @uwinx, is not needed. API can generate this values only in lowercase

JrooTJunior:
I don't like "Extensions" title in docs but accept this :D

PR #356, opened 2020-06-14
feat: add ChatType builtin filter

Birdi7:
# Description

Add the ChatType filter. Old filters were marked as deprecated to retain backward compatibility.

The last filter from #151.

## Type of change

# How Has This Been Tested?

Just used an example 

# Checklist:


aiogram/dispatcher/filters/builtin.py
```diff
@@ -691,3 +691,17 @@ def __init__(self, is_forwarded: bool):
 
     async def check(self, message: Message):
         return bool(getattr(message, "forward_date")) is self.is_forwarded
+
+
+class ChatTypeFilter(BoundFilter):
+    key = 'chat_types'
+
+    def __init__(self, chat_types: typing.List[ChatType]):
+        self.chat_types = chat_types
```
uburuntu, line 700: Can you add support for single `ChatType`?

For example:
```python3
@dp.message_handler(chat_types=ChatType.PRIVATE)
```

Reference: `ContentTypeFilter` https://github.com/aiogram/aiogram/blob/006a5ef987b4bc59feb78b9ff3f54061f273f4b0/aiogram/dispatcher/filters/builtin.py#L459

Birdi7: In order to avoid confusion which we have with TextFilter, I decided not to add it. Instead, I think we should force users to use lists as input arguments in such filters

aiogram/types/chat.py
```diff
@@ -512,6 +513,7 @@ def _check(obj, chat_types) -> bool:
         return obj.type in chat_types
 
     @classmethod
+    @deprecated("This filter was moved to ChatTypeFilter, and will be removed in aiogram v3.0")
```
uwinx, line 516: What's the motivation of deprecating those `is_{type}` properties? I think we should just leave these features because they may remain useful as we don't get the result of filter's `check` in your current implementation.

aiogram/dispatcher/filters/builtin.py
```diff
@@ -691,3 +691,17 @@ def __init__(self, is_forwarded: bool):
 
     async def check(self, message: Message):
         return bool(getattr(message, "forward_date")) is self.is_forwarded
+
+
+class ChatTypeFilter(BoundFilter):
+    key = 'chat_types'
+
+    def __init__(self, chat_types: typing.List[ChatType]):
```
uwinx, line 699: why is `chat_types` type restricted to `typing.List[ChatType]`, why not `Set` or any other container?

I'd use `Container[ChatType]` and given [@uburuntu's commentary](https://github.com/aiogram/aiogram/pull/356/files#r442129735). To reflect the current API, I suppose it'd be nice if we could:

0. rename `chat_types` to a singular `chat_type`
1. handle single value passed as a string, in the end having small instance check like...
```python3
self.required_types: Set[str]
if isinstance(chat_type, str):
    self.required_types = {chat_type}
else:
    self.required_types = set(chat_type)
```

what do you think?

aiogram/types/chat.py
Birdi7: As far as I understood the context of this feature by @JrooTJunior [here](https://github.com/aiogram/aiogram/issues/151#issuecomment-513668019), current filters should be marked as deprecated.

By the way, `ChatType.is_private(obj)` is the same as `ChatTypeFilter(chat_types[ChatType.PRIVATE]).check(obj)`

aiogram/dispatcher/filters/builtin.py
Birdi7: Okay, yeap, I've implemented the single value passing. Take a look at https://github.com/aiogram/aiogram/pull/356/commits/768491d97604e9e25d53f432fe12c597121ee88e, please

aiogram/dispatcher/filters/builtin.py
Birdi7: implemented at https://github.com/aiogram/aiogram/pull/356/commits/768491d97604e9e25d53f432fe12c597121ee88e

aiogram/dispatcher/filters/builtin.py
```diff
@@ -694,14 +695,17 @@ async def check(self, message: Message):
...
+
+        self.chat_type: typing.Set[str] = set(chat_type)
 
     async def check(self, obj: Union[Message, CallbackQuery]):
         if isinstance(obj, Message):
             obj = obj.chat
         if isinstance(obj, CallbackQuery):
             obj = obj.message.chat
-        return obj.type in self.chat_types
+        return obj.type in self.chat_type
```
uwinx: suggestion if we respect obj delcared types (`Union[Message, CallbackQuery]`):
```python3
    async def check(self, obj: Union[Message, CallbackQuery]):
        if isinstance(obj, Message):
            chat = obj.chat
        elif isinstance(obj, CallbackQuery):
            chat = obj.message.chat
        else:
            return NotImplemented
        return chat.type in self.chat_type
```

Birdi7: yeap, sure! thank you
b9424106a71df0857e8456e6b7f77d9b0d6e1401

Birdi7: eventually, here is a warning, if input object doens't match the type hint
https://github.com/aiogram/aiogram/pull/356/commits/d98b4ba1811e506e60821de8c316edba65b04ece

aiogram/dispatcher/filters/builtin.py
uburuntu: @JrooTJunior, your thoughts? There has to be some kind of unified approach

aiogram/dispatcher/filters/builtin.py
```diff
@@ -691,3 +692,24 @@ def __init__(self, is_forwarded: bool):
...
+
+        self.chat_type: typing.Set[str] = set(chat_type)
+
+    async def check(self, obj: Union[Message, CallbackQuery]):
+        if isinstance(obj, Message):
+            obj = obj.chat
+        elif isinstance(obj, CallbackQuery):
+            obj = obj.message.chat
+        else:
+            logger.warning("ChatTypeFilter doesn't support %s as input", type(obj))
```
JrooTJunior, line 712: Is better to use `warnings.warn` here

uwinx: I'd do `assert isinstance(obj, (Message, CallbackQuery))` before `if/elif` and in else `return False`

Birdi7:
I don't really see any point in breaking the whole pipeline of filters. As
far as I remember, that's the reason why we didn't use `raise
NotImplementedError`

aiogram/dispatcher/filters/builtin.py
Birdi7: I don't really see any point in breaking the whole pipeline of filters. As far as I remember, that's the reason why we didn't use `raise NotImplementedError`. I'd like to keep the warning message in such case. Remember, that in other filters we don't have such way to handle unexpected input

Birdi7: fixed in https://github.com/aiogram/aiogram/pull/356/commits/0eb5a476fae8f28ed286db5e2df844aa31f57b14

Birdi7:
@JrooTJunior I fixed the merge conflicts, should we merge the PR?

JrooTJunior:
Yep

MasterGroosha:
Why `SUPER_GROUP` and not `SUPERGROUP` ?

examples/chat_types_filter.py
```diff
@@ -0,0 +1,30 @@
...
+
+# Configure logging
+logging.basicConfig(level=logging.INFO)
+
+# Initialize bot and dispatcher
+bot = Bot(token=API_TOKEN)
+dp = Dispatcher(bot)
+
+
+@dp.message_handler(chat_types=[ChatType.PRIVATE, ChatType.CHANNEL])
```
kamikazebr: Typo here, correct to use is chat_type= without 's'

Birdi7: You’ve viewed the outdated version. Yes, the correct one is `chat_type`

kamikazebr: I see. Btw in other part of code too not work. I'll find and mark you there or open issue.

PR #363, opened 2020-06-21
Fix markdown escaping issues

### LINKED ISSUE #360: markdown.escape_md() doesn't escape '=' sign
## Context

From the [docs](https://core.telegram.org/bots/api#markdownv2-style):

> In all other places characters '_', '*', '[', ']', '(', ')', '~', '`', '>', '#', '+', '-', '=', '|', '{', '}', '.', '!' must be escaped with the preceding character '\'.

* aiogram version: 2.9.2

## Expected Behavior

markdown.escape_md() escapes '=' sign

## Current Behavior

'=' sign is not escaped

trying to send such text results in the error:

> aiogram.utils.exceptions.CantParseEntities: Can't parse entities: character '=' is reserved and must be escaped with the preceding '\'

unintended:
# Description
Fixes #360 and issue when you have pre-escaped like text

## Type of change


tests/test_utils/test_markdown.py
```diff
@@ -0,0 +1,11 @@
+import pytest
+
+from aiogram.utils import markdown
+
+
+class Test_escape:
```
JrooTJunior, line 6:
```suggestion
class TestMarkdownEscape:
```

PR #383, opened 2020-07-12
Command filter with caption check

uburuntu:
# Command filter with caption check

Add possibility to pass commands in captions: photo, video, audio, etc. This change doesnt break backward compatibility.

Example usage:
```python3
@dp.message_handler(commands=['myCommand'], commands_ignore_caption=False, content_types=ContentType.ANY)
@dp.message_handler(Command(['myCommand'], ignore_caption=False), content_types=[ContentType.TEXT, ContentType.DOCUMENT])
```

## Type of change

New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

Tested on own bot in many cases.

### DIFF
--- a/aiogram/dispatcher/filters/builtin.py
+++ b/aiogram/dispatcher/filters/builtin.py
@@ -37,7 +37,8 @@ class Command(Filter):
     def __init__(self, commands: Union[Iterable, str],
                  prefixes: Union[Iterable, str] = '/',
                  ignore_case: bool = True,
-                 ignore_mention: bool = False):
+                 ignore_mention: bool = False,
+                 ignore_caption: bool = True):
         """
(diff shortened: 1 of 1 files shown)


aiogram/dispatcher/filters/builtin.py
```diff
@@ -105,7 +120,7 @@ async def check_command(message: types.Message, commands, prefixes, ignore_case=
         if (command.lower() if ignore_case else command) not in commands:
             return False
 
-        return {'command': Command.CommandObj(command=command, prefix=prefix, mention=mention)}
```
JrooTJunior, line 108: Why replaced by directly class usage instead of cls?

uburuntu: Looks better :) 
Because `classmethod` allows easy renaming and other refactroings.

PR #393, opened 2020-07-27
Add missed ChatPermissions to __all__

uburuntu:
# Add missed ChatPermissions to __all__

Fix for this warning:

Also standardize all `__all__`'s as sorted tuples along the project.

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

On my bot


uwinx:
> Also standardize all __all__'s as sorted tuples along the project.

Any links to this standard? As far as I know, it must be list of strings.

uburuntu:
https://docs.python.org/3/reference/simple_stmts.html#the-import-statement

> The public names defined by a module are determined by checking the module’s namespace for a variable named `__all__`; if defined, **it must be a sequence of strings** which are names defined or imported by that module.

It doesn't matter that they are lists or tuples. But tuple was already used in `types/`, so I decided use it everywhere for increasing consistency.

JrooTJunior (approved):

PR #396, opened 2020-07-30
dev-2.x

TrDex:
some typing notations' updates


uburuntu:
About: `data: typing.Optional[dict] = None` in arguments.

> If necessary, Optional[t] is added for function and method annotations if a default value equal to None is set.
https://docs.python.org/3/library/typing.html#typing.get_type_hints

IMO, 
```python3
bucket: typing.Dict = None, **kwargs):
```
better than
```python3
bucket: typing.Optional[dict] = None, **kwargs):
```

---

P.S. And 
```python3
bucket: dict = None, **kwargs):
```
better than
```python3
bucket: typing.Dict = None, **kwargs):
```

TrDex:
Actually, [PEP 484](https://www.python.org/dev/peps/pep-0484/#id29) says that
> A past version of this PEP allowed type checkers to assume an optional type when the default value is None, as in this code:
> 
> `def handle_employee(e: Employee = None): ...`
> 
> This would have been treated as equivalent to:
> 
> `def handle_employee(e: Optional[Employee] = None) -> None: ...`
> This is no longer the recommended behavior. Type checkers should move towards requiring the optional type to be made explicit.

And `dict` is better than `typing.Dict` because of [PEP 585](https://www.python.org/dev/peps/pep-0585/) that adds Type Hinting Generics in Python 3.9.

But anyway, my main complaint is that `FSMContext.get_data()` (as well as `MemoryStorage.get_data()`) method should take `default` keyword argument as `Optional[dict]` instead of `Optional[str]` since `get_data()` returns a dictionary.

PR #399, opened 2020-08-08
Custom routes to Aiohttp web server.

dementiev27:
Added opportunity to add custom routes to Aiohttp web server while setting webhook.

Edited `start_webhook` method, now it takes `custom_routes` parameter (nested list of lists, e.g.: `[['GET', '/path', handler], ['POST', '/path2', handler_2]]`).

### DIFF
--- a/aiogram/utils/executor.py
+++ b/aiogram/utils/executor.py
@@ -2,7 +2,7 @@ import asyncio
 import datetime
 import functools
 import secrets
-from typing import Callable, Union, Optional, Any
+from typing import Callable, Union, Optional, Any, List, Coroutine
 from warnings import warn
 
 from aiohttp import web
--- a/aiogram/utils/executor.py
+++ b/aiogram/utils/executor.py
@@ -69,9 +69,9 @@ def set_webhook(dispatcher: Dispatcher, webhook_path: str, *, loop: Optional[asy
     return executor
 
 
-def start_webhook(dispatcher, webhook_path, *, loop=None, skip_updates=None,
-                  on_startup=None, on_shutdown=None, check_ip=False, retry_after=None, route_name=DEFAULT_ROUTE_NAME,
-                  **kwargs):
+def start_webhook(dispatcher, webhook_path, *, loop=None, skip_updates=None, on_startup=None, on_shutdown=None,
+                  check_ip=False, retry_after=None, route_name=DEFAULT_ROUTE_NAME,
+                  custom_routes: List[List[Union[str, str, Coroutine[Any, Any, Response]]]] = None, **kwargs):
     """
     Start bot in webhook mode
 
@@ -83,6 +83,7 @@ def start_webhook(dispatcher, webhook_path, *, loop=None, skip_updates=None,
     :param on_shutdown:
     :param check_ip:
     :param route_name:
+    :param custom_routes:
     :param kwargs:
     :return:
     """
@@ -95,8 +96,12 @@ def start_webhook(dispatcher, webhook_path, *, loop=None, skip_updates=None,
                            check_ip=check_ip,
                            retry_after=retry_after,
                            route_name=route_name)
-    executor.run_app(**kwargs)
 
+    if custom_routes != None:
+        for route in custom_routes:
+            executor.web_app.router.add_route(*route)
+
+    executor.run_app(**kwargs)
 
 def start(dispatcher, future, *, loop=None, skip_updates=None,
           on_startup=None, on_shutdown=None):
--- a/aiogram/utils/executor.py
+++ b/aiogram/utils/executor.py
@@ -97,7 +97,7 @@ def start_webhook(dispatcher, webhook_path, *, loop=None, skip_updates=None, on_
                            retry_after=retry_after,
                            route_name=route_name)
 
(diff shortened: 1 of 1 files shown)


gabbhack:
I don't think adding this to the telegram bot framework makes sense. You can just write a web server using aiohttp and then add a [WebhookRequestHandler](https://github.com/aiogram/aiogram/blob/22094eb477711e644924eaeb2424c8cef20e637a/aiogram/dispatcher/webhook.py#L75) to app.

dementiev27:
It might be really useful when you need to get webhooks from other services like payment systems etc.
Also, it's much easier and neatly than

> just write a web server using aiohttp and then add a [WebhookRequestHandler](https://github.com/aiogram/aiogram/blob/22094eb477711e644924eaeb2424c8cef20e637a/aiogram/dispatcher/webhook.py#L75) to app.

aiogram/utils/executor.py
```diff
@@ -69,9 +69,9 @@ def set_webhook(dispatcher: Dispatcher, webhook_path: str, *, loop: Optional[asy
     return executor
 
 
-def start_webhook(dispatcher, webhook_path, *, loop=None, skip_updates=None,
-                  on_startup=None, on_shutdown=None, check_ip=False, retry_after=None, route_name=DEFAULT_ROUTE_NAME,
-                  **kwargs):
+def start_webhook(dispatcher, webhook_path, *, loop=None, skip_updates=None, on_startup=None, on_shutdown=None,
+                  check_ip=False, retry_after=None, route_name=DEFAULT_ROUTE_NAME,
+                  custom_routes: List[List[Union[str, str, Coroutine[Any, Any, Response]]]] = None, **kwargs):
```
uwinx, line 74: strange type, don't you find?

dementiev27: It seems to me, this is the correct type-hinting for the nested list of lists of the union: method (str), path (str), and handler, isn't it?

dementiev27: This type-hinting can be removed, as you wish.

uwinx:
@dementiev27, thank you for the feature request.

I'm -1 for this as of current implementation and I don't think this should be library's responsibility.

JrooTJunior:
I don't like this change and this change does not correspond to the ideology of aiogram.

Executor is designed for simple cases when you only need to start dispatching events in webhook or polling mode without something else like integration with other services or etc.

So if you need to do something else you will need to create own instance of aiohttp application with adding request handler(s) and own logic. In this case you can use all aiohttp guides and attach WebhookRequestHandler to you own application or make something like you own WebhookRequestHandler.

Also you can use another web frameworks with aiogram for listening telegram webhook events like FastAPI or Sanic or something else and in this case aiogram don't have out of the box ready to use web handlers but you can to this.

I will close this PR. Sorry about that.

dementiev27:
@JrooTJunior, thanks for your reply!

aiogram/utils/executor.py
uwinx: it's about correctness of your type-hint.

`List[List[Union[str, str, Coroutine[Any, Any, Response]]]]`

- First thing to look at is `Union[str, str]` which always gives `str`. [docs](https://docs.python.org/3/library/typing.html#typing.Union)

- Secondly, `var = List[Union[str, str, X]]` narrows to `var = List[Union[str, X]]` and what do we get from the type-hint now? that `var` is a list of either string or X or both. but if we want...
> method (str), path (str), and handler
we probably want to use `List[Tuple[X...]]` instead of `List[List[Union[X...]]]`

- Lastly, why are we waiting for `Coroutine[Any, Any, Response]` for `handler` param?
I'm sure it's not what aiohttp really expects. What it really does expect from you is `Callable[[Request], Awaitable[StreamResponse]]` or `AbstractView`

Let me know, if anything was unclear or incorrect in my comment. typing is sure a great topic to discuss :)

dementiev27: Oh, you're completely right.
I realized that I was using Union wrong.
What about the last remark, actually it used to be something like this:

> Callable[[Request], Awaitable[StreamResponse]] or AbstractView

but PyCharm showed a warning like _expected ..., got `Coroutine[Any, Any, Response]` instead._

Thanks for your critics, since I'm a beginner I really appreciate it.

dementiev27:
Sorry for bothering you, but maybe you could help me with 

> You can just write a web server using aiohttp and then add a [WebhookRequestHandler](https://github.com/aiogram/aiogram/blob/22094eb477711e644924eaeb2424c8cef20e637a/aiogram/dispatcher/webhook.py#L75) to app.

I don't know how to add a `WebhookRequestHandler` to my aiohttp server.
I would be extremely happy if you could provide any examples of it.

Thank you.

JrooTJunior:
> I would be extremely happy if you could provide any examples of it.

Look at `aiogram` code:
https://github.com/aiogram/aiogram/blob/22094eb477711e644924eaeb2424c8cef20e637a/aiogram/dispatcher/webhook.py#L283-L294
https://github.com/aiogram/aiogram/blob/22094eb477711e644924eaeb2424c8cef20e637a/aiogram/utils/executor.py#L222-L261

PR #401, opened 2020-08-11
Command filter with forwarded messages check

uburuntu:
# Command filter with caption check

Add possibility to skip forwarded messages. This change does not break backward compatibility.

Example usage:
```python3
@dp.message_handler(commands=['my_command'], commands_ignore_forward=True)
```

_Note_: this PR should be merged after https://github.com/aiogram/aiogram/pull/390.

## Type of change

New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

Tested on own bot.


uburuntu:
Hmm, I think this possibility should be as another filter.

PR #409, opened 2020-08-29
Fix HTML characters escaping

WhiteMemory99:
# Description

html.escape replaces " and ' characters by default which is **NOT** required by Telegram and causes unexpected behavior with single quotes. `quote=False` argument fixes that.
image

## Type of change

# Checklist:


Latand (commented):
I confirm this problem
image

JrooTJunior (approved):

PR #414, opened 2020-09-03
Fix #413 parse entities positioning

### LINKED ISSUE #413: Broken text length and entity positions for a message that contain text with emojis
## Context

* Operating System:
* Python Version: 3.7.7
* aiogram version: 2.9.2

if you send a message "🚀**Hello Hello**" to the bot and call message.parse_entities(False), you will get `🚀H*ello Hello*`

## Expected Behavior

message.parse_entities(False) returns `🚀*Hello Hello*`

## Current Behavior

message.parse_entities(False) returns `🚀H*ello Hello*`

The problem is in json conversion and decoding of emoji surrogates. I debugged the protocol and getUpdates method returns the following json:
`{"ok":true,"result":[{"update_id":337194970,"message":{"message_id":12146,"from":{"id":535059,"is_bot":false,"first_name":"\u0418\u043b\u044c\u044f","username":"igonzo","language_code":"en"},"chat":{"id":535059,"first_name":"\u0418\u043b\u044c\u044f","username":"igonzo","type":"private"},"date":1599124626,"text":"\ud83d\ude80Hello Hello","entities":[{"offset":2,"length":11,"type":"bold"}]}}]}`
So the "text" has length 13 chars. And the bold entity starts at second char (offset=2) because emoji is encoded by two symbols (`\ud83d\ude80`). However after all the transformations the message object contains text of 12 unicode chars (because `\ud83d\ude80` are joined into one 🚀char), but the entities are still 13-length based. After you try to apply the entities to the text (e.g. as it's done in message.parse_entities()), you get a wrong positioning.

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.
1. Run the bot with the following code:

```
import os

from aiogram import Bot, Dispatcher, executor, types
from aiogram.types import ContentType

bot = Bot(token=os.environ['BOT_TOKEN'])
dp = Dispatcher(bot)


@dp.message_handler(content_types=ContentType.ANY)
async def demarkdown(message: types.Message):
  await message.reply(message.parse_entities(False),
                      disable_web_page_preview=True)


if __name__ == '__main__':
  executor.start_polling(dp)

```
2. Send "🚀**Hello Hello**" to the bot (formatted)
3. You get `🚀H*ello Hello*` replied

unintended:
# Description

Fixes #413

## Type of change

### DIFF
--- a/aiogram/utils/text_decorations.py
+++ b/aiogram/utils/text_decorations.py
@@ -56,15 +56,9 @@ class TextDecoration(ABC):
         :return:
         """
 
-        text = text.encode('utf-16-le')
-        entities = [MessageEntity(**e.to_python()) for e in (entities if entities else [])]
-        for e in entities:
-            e.offset *= 2
-            e.length *= 2
-
         result = "".join(
             self._unparse_entities(
-                text, sorted(entities, key=lambda item: item.offset) if entities else []
+                text.encode('utf-16-le'), sorted(entities, key=lambda item: item.offset) if entities else []
             )
         )
         return result
@@ -81,15 +75,15 @@ class TextDecoration(ABC):
         length = length or len(text)
 
         for index, entity in enumerate(entities):
-            if entity.offset < offset:
+            if entity.offset * 2 < offset:
                 continue
-            if entity.offset > offset:
-                yield self.quote(text[offset : entity.offset].decode('utf-16-le'))
-            start = entity.offset
-            offset = entity.offset + entity.length
+            if entity.offset * 2> offset:
+                yield self.quote(text[offset : entity.offset * 2].decode('utf-16-le'))
+            start = entity.offset * 2
+            offset = entity.offset * 2 + entity.length * 2
 
             sub_entities = list(
-                filter(lambda e: e.offset < (offset or 0), entities[index + 1 :])
+                filter(lambda e: e.offset * 2 < (offset or 0), entities[index + 1 :])
             )
             yield self.apply_entity(
                 entity,
--- a/aiogram/utils/text_decorations.py
+++ b/aiogram/utils/text_decorations.py
@@ -55,6 +55,13 @@ class TextDecoration(ABC):
         :param entities: Array of MessageEntities
         :return:
         """
+
+        text = text.encode('utf-16-le')
+        entities = [MessageEntity(**e.to_python()) for e in (entities if entities else [])]
+        for e in entities:
+            e.offset *= 2
+            e.length *= 2
+
         result = "".join(
             self._unparse_entities(
                 text, sorted(entities, key=lambda item: item.offset) if entities else []
@@ -64,7 +71,7 @@ class TextDecoration(ABC):
 
     def _unparse_entities(
         self,
-        text: str,
+        text: bytes,
         entities: List[MessageEntity],
         offset: Optional[int] = None,
         length: Optional[int] = None,
@@ -77,7 +84,7 @@ class TextDecoration(ABC):
             if entity.offset < offset:
                 continue
             if entity.offset > offset:
-                yield self.quote(text[offset : entity.offset])
+                yield self.quote(text[offset : entity.offset].decode('utf-16-le'))
             start = entity.offset
             offset = entity.offset + entity.length
 
@@ -94,7 +101,7 @@ class TextDecoration(ABC):
             )
 
         if offset < length:
-            yield self.quote(text[offset:length])
+            yield self.quote(text[offset:length].decode('utf-16-le'))
 
     @abstractmethod
     def link(self, value: str, link: str) -> str:  # pragma: no cover
--- a/aiogram/utils/text_decorations.py
+++ b/aiogram/utils/text_decorations.py
@@ -55,7 +55,6 @@ class TextDecoration(ABC):
         :param entities: Array of MessageEntities
         :return:
         """
-
         result = "".join(
             self._unparse_entities(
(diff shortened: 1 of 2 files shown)


JrooTJunior:
LGTM

Can you also make backport to `dev-3.x` branch with adding corresponding test that prove you changes?


And that is something strange. When i check entity length after adding MarkdownV2 i didn't see the need to use `utf-16-le` and then i remove it. So as i an see in your description i was wrong... Thanks

aiogram/utils/text_decorations.py
```diff
@@ -57,14 +57,14 @@ def unparse(self, text: str, entities: Optional[List[MessageEntity]] = None) ->
         """
         result = "".join(
             self._unparse_entities(
-                text, sorted(entities, key=lambda item: item.offset) if entities else []
+                text.encode('utf-16-le'), sorted(entities, key=lambda item: item.offset) if entities else []
```
JrooTJunior, line 60: Small optimization.

Please move calling of encode to separated method:
```python3
def _add_surrogates(self, text: str) -> bytes:
    return text.encode('utf-16-le')
``` 
And also use adding/removing surrogates inside `_unparse_entities` method

aiogram/utils/text_decorations.py
```diff
@@ -57,14 +57,14 @@ def unparse(self, text: str, entities: Optional[List[MessageEntity]] = None) ->
...
             )
         )
         return result
 
     def _unparse_entities(
         self,
-        text: str,
+        text: bytes,
```
JrooTJunior, line 67: I think text should be passed to this function without any changes and then encoding should be placed inside this method

aiogram/utils/text_decorations.py
```diff
@@ -74,15 +74,15 @@ def _unparse_entities(
...
+            if entity.offset * 2 < offset:
                 continue
-            if entity.offset > offset:
-                yield self.quote(text[offset : entity.offset])
-            start = entity.offset
-            offset = entity.offset + entity.length
+            if entity.offset * 2> offset:
+                yield self.quote(text[offset : entity.offset * 2].decode('utf-16-le'))
+            start = entity.offset * 2
+            offset = entity.offset * 2 + entity.length * 2
```
JrooTJunior, line 82: Very small optimization:
```suggestion
            offset = (entity.offset + entity.length) * 2
```

Or maybe calculate doubled size once and then use it

For example:
```python3
for index, entity in enumerate(entities):
    entity_length = entity.length * 2
    entity_offset = entity.offset * 2


    if entity_offset < offset:
        continue
    ...
```

aiogram/utils/text_decorations.py
```diff
@@ -74,15 +74,15 @@ def _unparse_entities(
...
         for index, entity in enumerate(entities):
-            if entity.offset < offset:
+            if entity.offset * 2 < offset:
                 continue
-            if entity.offset > offset:
-                yield self.quote(text[offset : entity.offset])
-            start = entity.offset
-            offset = entity.offset + entity.length
+            if entity.offset * 2> offset:
+                yield self.quote(text[offset : entity.offset * 2].decode('utf-16-le'))
```
JrooTJunior, line 80: Move decoding to separated method:

```python3
def _remove_surrogates(self, data: bytes) -> str:
    return data.decode('utf-16-le')
```

And then use it here as
```python3
yield self.quote(self._remove_surrogates(text[offset : entity_offset]))
```

aiogram/utils/text_decorations.py
```diff
@@ -94,7 +94,7 @@ def _unparse_entities(
             )
 
         if offset < length:
-            yield self.quote(text[offset:length])
+            yield self.quote(text[offset:length].decode('utf-16-le'))
```
JrooTJunior, line 97:
```python3
yield self.quote(self._remove_surrogates(text[offset:length]))
```

JrooTJunior (changes requested):

aiogram/utils/text_decorations.py
unintended: I was trying to do that way, but there is recursive call of _unparse_entities, so it may lead to unexpected results for nested entities. I didn't have time to check the behavior and tine this index magic, so I decided to move decode/encode outside and pass the bytes. That's why I called the solution "raw" :)

JrooTJunior: Hm.. Just try it and if this way is hard leave it as is 🙂

JrooTJunior: And with recursive call, hmm, encoding outside `_unparse_entities` make sense..

unintended:
I updated the code: added tests, refactored encoding to methods, but didn't manage to put surrogates methods inside `_unparse_entities`. Maybe someone will try to make make it one day :) 
Also tried to port to dev-3.x, but with no luck yet (couldn't port the tests properly becoase of different nature of api classes) and I don't have much time to do it, I'm sorry.

JrooTJunior:
in `dev-3.x` you can add test-case to parameterized test of decorations:
https://github.com/aiogram/aiogram/blob/6f53f15577e3b05a0268b78efdfb410e95672429/tests/test_utils/test_text_decorations.py#L110-L231 

and at the end of list add item:
```python3
[
    markdown_decoration,
    "🚀 i'm bold and italic and still bold",
    [
        MessageEntity(offset=3, length=34, type="bold"),
        MessageEntity(offset=12, length=10, type="italic"),
    ],
    "🚀 *i'm bold _and italic_\r and still bold*"
]
```

JrooTJunior (approved):

JrooTJunior:
Thanks for contribution :wink:

PR #424, opened 2020-09-14
hotfix(dispatcher): when DP.loop None

uwinx:
property DP.loop may be None, executor and webhook view should both respect that


Olegt0rr (approved):
Executor checked. It's okay

- [x] polling
- [ ] webhook

JrooTJunior (approved):

PR #435, opened 2020-10-03
HandlerObject: Unwrap handler before awaitable check

### LINKED ISSUE #434: 3.x: HandlerObject doesn't consider decorated class handlers as awaitable
## Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Python Version: 3.9.0b3
* aiogram version: 3.0.0a5
* aiohttp version: 3.6.2
* uvloop version (if installed): 

## Expected Behavior

await handlers :P

## Current Behavior

Handlers wont get awaited

When using wrapped decorator over the class handlers, `HandlerObject` conclude it doesn't need to awaited. I debugged a little and find out that when its not class handlers and just functions, `CallableMixin` unwraps the function.
https://github.com/aiogram/aiogram/blob/6f53f15577e3b05a0268b78efdfb410e95672429/aiogram/dispatcher/event/handler.py#L25

When it comes to class handlers, aiogram doesn't unwrap so callable is just a wrapped function and skips this check 
https://github.com/aiogram/aiogram/blob/6f53f15577e3b05a0268b78efdfb410e95672429/aiogram/dispatcher/event/handler.py#L62-L66

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

Just run this

```python
def test_decorator(func):
    @wraps(func)
    async def wrapped(*args, **kwargs):
        return await func(*args, **kwargs)
    return wrapped

@router.message(...)
@test_decorator
class Test(MesssageHandler):
    async def handle(self):
        pass

```

### Failure Logs

Please include any relevant log snippets or files here.

sabbyX:
# Description

Unwraps the callback before awaitable check

Fixes #434 

## Type of change

# Checklist:

### DIFF
--- a/aiogram/dispatcher/event/handler.py
+++ b/aiogram/dispatcher/event/handler.py
@@ -62,7 +62,7 @@ class HandlerObject(CallableMixin):
     def __post_init__(self) -> None:
         super(HandlerObject, self).__post_init__()
         callback = inspect.unwrap(self.callback)
-        if inspect.isclass(callback) and issubclass(callback, BaseHandler):  # type: ignore
+        if inspect.isclass(callback) and issubclass(callback, BaseHandler):
             self.awaitable = True
 
     async def check(self, *args: Any, **kwargs: Any) -> Tuple[bool, Dict[str, Any]]:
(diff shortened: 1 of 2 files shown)


JrooTJunior:
Thanks!
Can you add tests to prove this fix?

As can i see this change doesn't affect or break the current tests but what about covering problem from issue #434?

sabbyX:
> Thanks!
> Can you add tests to prove this fix?
> 
> As can i see this change doesn't affect or break the current tests but what about covering problem from issue #434?

Done

JrooTJunior (approved):

PR #442, opened 2020-10-13
feat(InputFile): adding Pathlib to supporting types for files

amalkhamidov:
# Description

❤️ This change allows the Pathlib library to be used as a path reference for InputFile!

### DIFF
--- a/aiogram/types/input_file.py
+++ b/aiogram/types/input_file.py
@@ -51,7 +51,8 @@ class InputFile(base.TelegramObject):
         elif isinstance(path_or_bytesio, Path):
             self._file = path_or_bytesio.open("rb")
             self._path = path_or_bytesio.resolve()
-
+            if filename is None:
+                filename = path_or_bytesio.name
         else:
             raise TypeError('Not supported file type.')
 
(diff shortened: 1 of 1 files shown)


evgfilim1 (commented):
LGTM, thanks

aiogram/types/input_file.py
```diff
@@ -45,6 +47,11 @@ def __init__(self, path_or_bytesio, filename=None, conf=None):
         elif isinstance(path_or_bytesio, _WebPipe):
             self._path = None
             self._file = path_or_bytesio
+
+        elif isinstance(path_or_bytesio, Path):
+            self._file = path_or_bytesio.open("rb")
+            self._path = path_or_bytesio.resolve()
+
```
JrooTJunior, line 54: Filename should be resolved in the same way as it with string `path_or_bytesio`

```suggestion
            if filename is None:
                filename = path_or_bytesio.name
```

JrooTJunior (changes requested):

JrooTJunior (approved):

PR #387, opened 2020-07-21
fix(aiohttp-session): respect UNSET sentinel

uwinx:
# Description

check if value is `UNSET` while building http request.

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

PR #450, opened 2020-10-31
Fix unsupported parse_mode

evgfilim1:
## Summary
Fixed `Bad Request: unsupported parse_mode` error due to missing `prepare_parse_mode` line

P.S. Do I have to cover this case by tests?

UPD. Might be a bit related to #387


JrooTJunior:
Covered by #469

PR #453, opened 2020-11-04
code beautify

muhammedfurkan:


aiogram/types/inline_keyboard.py
```diff
@@ -50,7 +50,7 @@ def add(self, *args):
             if index % self.row_width == 0:
                 self.inline_keyboard.append(row)
                 row = []
-        if len(row) > 0:
+        if row:
```
evgfilim1, line 53: Explicit is better than implicit

aiogram/types/reply_keyboard.py
```diff
@@ -58,7 +58,7 @@ def add(self, *args):
             if index % self.row_width == 0:
                 self.keyboard.append(row)
                 row = []
-        if len(row) > 0:
+        if row:
```
evgfilim1, line 61: Explicit is better than implicit

tests/test_utils/test_deprecated.py
```diff
@@ -7,8 +7,8 @@ def test_DeprecatedReadOnlyClassVarCD():
     assert DeprecatedReadOnlyClassVar.__slots__ == ("_new_value_getter", "_warning_message")
 
     new_value_of_deprecated_cls_cd = "mpa"
-    pseudo_owner_cls = type("OpekaCla$$", (), {})
```
evgfilim1, line 10: Why have you moved it?

aiogram/types/inline_keyboard.py
uburuntu: https://google.github.io/styleguide/pyguide.html#2144-decision

aiogram/types/fields.py
```diff
@@ -126,42 +126,32 @@ def __init__(self, *args, **kwargs):
         if default is None:
             default = []
 
-        super(ListField, self).__init__(*args, default=default, **kwargs)
+        super().__init__(*args, default=default, **kwargs)
```
Olegt0rr, line 129: I wanted to suggest, but it was applied 0_o

aiogram/utils/helper.py
```diff
@@ -201,10 +194,15 @@ class OrderedHelperMeta(type):
...
+        props_keys = [
+            prop_name
+            for prop_name in (
+                name
+                for name, prop in namespace.items()
+                if isinstance(prop, (Item, ListItem))
+            )
+        ]
 
-        for prop_name in (name for name, prop in namespace.items() if isinstance(prop, (Item, ListItem))):
```
Olegt0rr, line 206: Seems it's more readable

JrooTJunior:
OMG :scream: 

That is... is.. is.. painful pull-request... I'm shocked...

Before start.. black, isort can't be applied to `dev-2.x` branch in due to this change breaks all opened pull-request and makes it impossible. I't breaks everything.

After partial reviewing you'r changes - tests was changed absolutely incorrect and code-style doesn't follows applied by our team rules.

For example:
```python
if <condition>:
    raise AssertionError
```
absolutelly incorrect in use instead of 
```python
assert <condition>
```
this change makes sense only when code will be started with `-O` interpreter option but not fully.

---

Also we don't planned to enable any automated code-checks for `dev-2.x` branch because already initialized development of `3.0` version where is enabled CI/CD with black, isort, flake8, mypy checks in strict mode. So if you want to make code prettified - lets look at `dev-3.x` branch.

---

I don't like to make negative or aggressive comments on issues/PR but... Changes should be applied with common sense and be adequate but you're changes are not adequate...

Issue #405, opened 2020-08-15
setdefault in FSMContextProxy doesn't return the proper value


Ibraheem-Asaad:
## Expected Behavior

setdefault returning the value of the provided key if found or the provided default if not (the same as dict's builtin behavior)

## Current Behavior

The corresponding value isn't returned (None is returned as default)

## Proposed Solution

Simply returning the value in this line https://github.com/aiogram/aiogram/blob/master/aiogram/dispatcher/storage.py#L400

Ibraheem-Asaad:
Fix was merged

PR #486, opened 2020-12-29
fix setdefault of FSMContextProxy

OGURCHINSKIY:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.

Fixes # (issue)

## Type of change

Please delete options that are not relevant.

- [ ] Documentation (typos, code examples or any documentation update)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Operating System: 
* Python version: 

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #491, opened 2021-01-09
fix return value of FSMContextProxy's setdefault (#405)

Ibraheem-Asaad:
# Description

setdefault to actually return the proper value

Fixes issue #405

## Type of change

Bug fix (non-breaking change which fixes an issue)


JrooTJunior:
Duplicates #486. But author of #486 did not resolves an code-review messages. So that's mean this PR should be merged.

JrooTJunior (approved):

PR #497, opened 2021-01-12
Add ability to `.answer_*` and `.reply_*` live location

TrDex:
# Description

Added the ability to pass more parameters (`kwargs`) from methods `Message().answer_location()` and `Message().reply_location()` to `self.bot.send_location()`

## Type of change

# How Has This Been Tested?

No tests run. I believe it's a small change and it doesn't break anything.

# Checklist:

### DIFF
--- a/aiogram/types/message.py
+++ b/aiogram/types/message.py
@@ -957,10 +957,7 @@ class Message(base.TelegramObject):
         self,
         latitude: base.Float,
         longitude: base.Float,
-        horizontal_accuracy: typing.Optional[base.Float] = None,
         live_period: typing.Optional[base.Integer] = None,
-        heading: typing.Optional[base.Integer] = None,
-        proximity_alert_radius: typing.Optional[base.Integer] = None,
         disable_notification: typing.Optional[base.Boolean] = None,
         allow_sending_without_reply: typing.Optional[base.Boolean] = None,
         reply_markup: typing.Union[
@@ -971,6 +968,9 @@ class Message(base.TelegramObject):
             None,
         ] = None,
         reply: base.Boolean = False,
+        horizontal_accuracy: typing.Optional[base.Float] = None,
+        heading: typing.Optional[base.Integer] = None,
+        proximity_alert_radius: typing.Optional[base.Integer] = None,
     ) -> Message:
         """
         Use this method to send point on the map.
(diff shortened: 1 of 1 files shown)


aiogram/types/message.py
```diff
@@ -957,7 +957,10 @@ async def answer_location(
         self,
         latitude: base.Float,
         longitude: base.Float,
+        horizontal_accuracy: typing.Optional[base.Float] = None,
         live_period: typing.Optional[base.Integer] = None,
+        heading: typing.Optional[base.Integer] = None,
+        proximity_alert_radius: typing.Optional[base.Integer] = None,
```
evgfilim1, line 963: This change will break a code like this:
```python3
await message.answer_location(42.0, 42.0, 60, True)
```
While originally it was assumed to be
```python3
await message.answer_location(latitude=42.0, longitude=42.0, live_period=60, disable_notification=True)
```
it will be (after applying this patch)
```python3
await message.answer_location(latitude=42.0, longitude=42.0, horizontal_accuracy=60, live_period=True)
```

@JrooTJunior what do you think about this?

aiogram/types/message.py
```diff
@@ -2055,7 +2074,10 @@ async def reply_location(
...
```
evgfilim1, line 2080: Same here

TrDex:
@evgfilim1 @JrooTJunior WDYT?

JrooTJunior (approved):

JrooTJunior:
LGTM

@evgfilim1, what do you think?

aiogram/types/message.py
```diff
@@ -968,6 +968,9 @@ async def answer_location(
             None,
         ] = None,
         reply: base.Boolean = False,
+        horizontal_accuracy: typing.Optional[base.Float] = None,
+        heading: typing.Optional[base.Integer] = None,
+        proximity_alert_radius: typing.Optional[base.Integer] = None,
```
evgfilim1, line 973: I think it'd be better to leave `reply` the last arg as it is not really related to the API but to our shortcuts. Also `allow_sending_without_reply` is API 5.0 change, so IMO we can really put these new args before `reply_markup` as I don't think there is anyone who does smth like `message.answer_location(42.0, 42.0, 60, True, False, keyboard, False)`, otherwise it's weird, isn't it? :smile: 

TL;DR I guess `reply_markup` is usually used as a keyword-arg, so new args would fit best before `reply_markup`

aiogram/types/message.py
```diff
@@ -2065,6 +2084,9 @@ async def reply_location(
             None,
         ] = None,
         reply: base.Boolean = True,
+        horizontal_accuracy: typing.Optional[base.Float] = None,
+        heading: typing.Optional[base.Integer] = None,
+        proximity_alert_radius: typing.Optional[base.Integer] = None,
```
evgfilim1, line 2089: Same here (don't forget to change if you're going to)

evgfilim1 (approved):
LGTM, thanks for contribution!

uwinx (approved):

PR #501, opened 2021-02-01
Added answer_chat_action() method for types.Message

klkvr:
Addded answer_chat_action() method for types.Message which allows to use send_chat_action() method directly from recieved message.


uwinx (approved):
lgtm

JrooTJunior (approved):

PR #509, opened 2021-02-20
Fix empty bucket return

PavelMinev:
# Description

When trying to implement throttling for commands "hr = await dispatcher.check_key(key)" without other parameters - it couldn't find any bucket.

## Type of change

# How Has This Been Tested?

- [x] Implementing a simple middleware to throttle commands. Trying to get current Throttled object from bucket with: `thr = await dispatcher.check_key(key)` returns an empty Throttled (zero and None values). Checking this code it was found, that it should be `user_id` and `chat_id` parameters passed to `storage.get_bucket` method.

# Checklist:


JrooTJunior:
Duplicate #520

PR #520, opened 2021-03-10
fix bug in getting user_id, chat_id from context

darksidecat:
in Dispatcher check_key and release_key functions
(need User.id for future use, not User object)

# Description

there was a bug in check_key and release_key functions.
When we call these functions with no chat_id or user_id argument, they cannot set user_id and chat id variable correctly, becouse this code:

user_id = types.User.get_current()
chat_id = types.Chat.get_current()

return User/Chat object, not user/chat id

## Type of change

# How Has This Been Tested?

- [x] Run aiogram tests
- [x] Run bot with these changes

**Test Configuration**:
* Operating System: Windows 10 | Linux Manjaro
* Python version: 3.7.10 |  3.9.2

# Checklist:

- [ ]  I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/aiogram/dispatcher/dispatcher.py
+++ b/aiogram/dispatcher/dispatcher.py
@@ -1136,8 +1136,8 @@ class Dispatcher(DataMixin, ContextInstanceMixin):
             raise RuntimeError('This storage does not provide Leaky Bucket')
 
         if user_id is None and chat_id is None:
-            user_id = types.User.get_current()
-            chat_id = types.Chat.get_current()
+            user_id = types.User.get_current().id
+            chat_id = types.Chat.get_current().id
 
         bucket = await self.storage.get_bucket(chat=chat_id, user=user_id)
         data = bucket.get(key, {})
@@ -1158,8 +1158,8 @@ class Dispatcher(DataMixin, ContextInstanceMixin):
             raise RuntimeError('This storage does not provide Leaky Bucket')
 
         if user_id is None and chat_id is None:
-            user_id = types.User.get_current()
-            chat_id = types.Chat.get_current()
+            user_id = types.User.get_current().id
+            chat_id = types.Chat.get_current().id
 
         bucket = await self.storage.get_bucket(chat=chat_id, user=user_id)
         if bucket and key in bucket:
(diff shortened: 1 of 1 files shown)


aiogram/dispatcher/dispatcher.py
```diff
@@ -1136,8 +1136,8 @@ async def check_key(self, key, chat_id=None, user_id=None):
             raise RuntimeError('This storage does not provide Leaky Bucket')
 
         if user_id is None and chat_id is None:
-            user_id = types.User.get_current()
-            chat_id = types.Chat.get_current()
+            user_id = types.User.get_current().id
```
Olegt0rr, line 1139: `get_current()` can return `None`, please check this case

aiogram/dispatcher/dispatcher.py
```diff
@@ -1158,8 +1158,8 @@ async def release_key(self, key, chat_id=None, user_id=None):
...
```
Olegt0rr, line 1161: `get_current()` can return `None`, please check this case

Olegt0rr (changes requested):

aiogram/dispatcher/dispatcher.py
darksidecat: Ok, I'll do this.
There is also User.get_current().id in throttle function, I'll make check this too.

darksidecat: Also, I think it would be better to split this condition 
`if user_id is None and chat_id is None:`

into two separate conditions in throttle, check_key, release_key functions
@Olegt0rr What do you think?

darksidecat: > Also, I think it would be better to split this condition
> `if user_id is None and chat_id is None:`
> 
> into two separate conditions in throttle, check_key, release_key functions
> @Olegt0rr What do you think?

maybe it's a bad idea, i need to think more about it,
if these functions will be called with user_id or chat_id argument

bucket will be managed for user_id == chat_id, can't decide if this behavior is always expected.
It looks like this is the expected behavior. And nothing needs to be changed

Olegt0rr: It's a paired id of bucket, so I think there's no reason to split it.
@JrooTJunior what do you think?

Olegt0rr (approved):

JrooTJunior (approved):

PR #542, opened 2021-03-25
Set state via storage

### LINKED ISSUE #541: States aren't being resolved in storage instances
## Context

FSMContext [resolves](https://github.com/aiogram/aiogram/blob/15a6e94135f86777af1ff96a7e40c1e5923558e3/aiogram/dispatcher/storage.py#L282) states if it's needed. However, neither BaseStorage, nor storages provide such method

* aiogram version: 2.12.1

## Current Behavior

```python
class Test(StatesGroup):
    test = State()

@dp.message_handler(commands="set_state")
async def set_state(message: types.Message, state: FSMContext):
    """In this case, the storage is RedisStorage2. Other storages,
    such as Memory and Mongo, behave the same way (kind of)"""

    await storage.set_state(user=message.from_user.id, state=Test.test)
    # Current state: None, Expected: Test:test
    
    await state.set_state(Test.test)
    # Works! Current state: Test:test
```

Olegt0rr:
# Description
Fixes #541

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Operating System: 
* Python version: 

# Checklist:

### DIFF
--- a/aiogram/dispatcher/storage.py
+++ b/aiogram/dispatcher/storage.py
@@ -273,15 +273,6 @@ class BaseStorage:
         """
         await self.set_data(chat=chat, user=user, data={})
 
-
-class FSMContext:
-    def __init__(self, storage, chat, user):
-        self.storage: BaseStorage = storage
-        self.chat, self.user = self.storage.check_address(chat=chat, user=user)
-
-    def proxy(self):
-        return FSMContextProxy(self)
-
     @staticmethod
     def resolve_state(value):
         from .filters.state import State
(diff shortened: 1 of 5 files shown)


aiogram/dispatcher/storage.py
```diff
@@ -270,29 +273,32 @@ async def reset_bucket(self, *,
...
     @staticmethod
-    def _resolve_state(value):
+    def resolve_state(value):
         from .filters.state import State
 
         if value is None:
             return
-        elif isinstance(value, str):
+
+        if isinstance(value, str):
```
uwinx, line 283: what's the purpose of creating new comparison branch for each type, when the instance in our case cannot be `State` and `str` at the same time?

uwinx (changes requested):

aiogram/dispatcher/storage.py
Olegt0rr: Since the return statement terminates the execution of the current function, the both forms (if+return/elif+return) are equivalent, but separate checks looks more readable

JrooTJunior (approved):

PR #546, opened 2021-03-27
i18n get_locale without User

### LINKED ISSUE #544: I18nMiddleware.get_user_locale() cant get user for some actions (Poll closing)
When user closing Poll we get update without information that can give us current User object, so we cant get user.locale

```python:
async def get_user_locale(self, action: str, args: Tuple[Any]) -> str:
    user: types.User = types.User.get_current()
    locale: Locale = user.locale <--- raised exception here "NoneType" object 
                                      has no attribute 'locale'
```

how it can be fixed:
* adding check for this `pre_process_poll` action to I18nMiddleware.trigger()
`and  'poll' not in action`

maybe need investigation for other actions

aiogram 2.12.1
python 3.9.2
Manjaro 21.0

Olegt0rr:
# Description
Fixes #544 

## Type of change
- [х] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?
- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Operating System: 
* Python version: 

# Checklist:

### DIFF
--- a/aiogram/contrib/middlewares/i18n.py
+++ b/aiogram/contrib/middlewares/i18n.py
@@ -136,6 +136,7 @@ class I18nMiddleware(BaseMiddleware):
             *_, data = args
             language = data['locale'] = locale.language
             return language
+        return None
 
     async def trigger(self, action, args):
         """
--- a/aiogram/contrib/middlewares/i18n.py
+++ b/aiogram/contrib/middlewares/i18n.py
@@ -1,7 +1,7 @@
 import gettext
 import os
 from contextvars import ContextVar
-from typing import Any, Dict, Tuple
+from typing import Any, Dict, Tuple, Optional
 
 from babel import Locale
 from babel.support import LazyProxy
@@ -129,7 +129,7 @@ class I18nMiddleware(BaseMiddleware):
         :return: locale name
         """
         user: types.User = types.User.get_current()
-        locale: Locale = user.locale if user else None
+        locale: Optional[Locale] = user.locale if user else None
 
         if locale:
             *_, data = args
--- a/aiogram/contrib/middlewares/i18n.py
+++ b/aiogram/contrib/middlewares/i18n.py
@@ -129,7 +129,7 @@ class I18nMiddleware(BaseMiddleware):
         :return: locale name
         """
         user: types.User = types.User.get_current()
-        locale: Locale = user.locale
+        locale: Locale = user.locale if user else None
 
         if locale:
             *_, data = args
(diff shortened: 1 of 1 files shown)


aiogram/contrib/middlewares/i18n.py
```diff
@@ -128,8 +128,8 @@ async def get_user_locale(self, action: str, args: Tuple[Any]) -> str:
         :param args: event arguments
         :return: locale name
```
uwinx, line 129: return in this function becomes `Optional[str]`

uwinx (changes requested):

aiogram/contrib/middlewares/i18n.py
```diff
@@ -122,7 +122,8 @@ def lazy_gettext(self, singular, plural=None, n=1, locale=None, enable_cache=Fal
     async def get_user_locale(self, action: str, args: Tuple[Any]) -> str:
```
uwinx:
```suggestion
    async def get_user_locale(self, action: str, args: Tuple[Any]) -> Optional[str]:
```

uwinx (changes requested):

aiogram/contrib/middlewares/i18n.py
Olegt0rr: Outdated code, already fixed
https://github.com/aiogram/aiogram/pull/546/files#diff-3eb6fb402a3689572f1a23bdb28610ddb0e2dcc5bd1ac4d3f5e24913706a5c16R122

aiogram/contrib/middlewares/i18n.py
```diff
@@ -119,22 +119,24 @@ def lazy_gettext(self, singular, plural=None, n=1, locale=None, enable_cache=Fal
...
-        user: types.User = types.User.get_current()
-        locale: Locale = user.locale
+        user: Optional[types.User] = types.User.get_current()
+        locale: Optional[Locale] = user.locale if user else None
 
         if locale:
             *_, data = args
             language = data['locale'] = locale.language
             return language
+        return None
```
Termonoid, line 139: When method returns None, await on line 152 fails.
`locale = await self.get_user_locale(action, args)` <- awaiting None occurs here

Termonoid: We should check for None in `trigger` before awaiting result

Termonoid (changes requested):

aiogram/contrib/middlewares/i18n.py
uwinx: Please elaborate. we `await` for a coroutine and not the None. the `get_user_locale(...)` may return a `Awaitable[None | str]`, not `None`.

JrooTJunior (approved):

PR #547, opened 2021-03-27
PEP8 dunders placement fix

Olegt0rr:
https://www.python.org/dev/peps/pep-0008/#module-level-dunder-names

> 
> Module level "dunders" (i.e. names with two leading and two trailing underscores) such as `__all__`, `__author__`, `__version__`, etc. should be placed after the module docstring but before any import statements except `from __future__ imports`. Python mandates that future-imports must appear in the module before any other code except docstrings:
> ```python
> """This is the example module.
> 
> This module does stuff.
> """
> 
> from __future__ import barry_as_FLUFL
> 
> __all__ = ['a', 'b', 'c']
> __version__ = '0.1'
> __author__ = 'Cardinal Biggles'
> 
> import os
> import sys
> ```

Also removed backslashes from imports
> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.


evgfilim1 (approved):

uwinx (approved):
LGTM.

However quite funny how we switch from one thing to another so fast, as of PR #393 it was introduced that we use tuples for `__all__` as a module-level variable.

Olegt0rr:
> However quite funny how we switch from one thing to another so fast, as of PR #393 it was introduced that we use tuples for `__all__` as a module-level variable.

Missed this discussion. 
All found references to `__all__`  from python docs mention this variable as a list.

* https://docs.python.org/3/tutorial/modules.html#importing-from-a-package
* https://www.python.org/dev/peps/pep-0008/#module-level-dunder-names

evgfilim1:
:thinking:

JrooTJunior:
v2 will no more be supported.

PR #558, opened 2021-04-06
fix: builtin command filter args (#556)

### LINKED ISSUE #556: 'args' attribute of 'Command.CommandObj' seems not working
## Context

* Python Version: 3.8.1
* aiogram version: 2.12.1
* aiohttp version: 3.7.3
* uvloop version (if installed): 

## Expected Behavior

`command.args` returns arguments of command

## Current Behavior

`command.args` returns `None`

It seems the `args` attribute is not working.

Digging into `dispatcher/filters/builtin.py`, I found nothing about setting this attribute, but after adding `args = text.split()[1]` and pass it to the construtor of `CommandObj`, it seems to work.

I wonder if this is the intended behavior or am I getting it wrong?

I would be glad to open a PR if it's a mistake :)

### Steps to Reproduce

```python
@dp.message_handler(commands=['test'])
async def test(message: Message, command: Command.CommandObj):
    print(command, command.args)
```

Test command: `/test abc`

### Failure Logs

Output: `Command.CommandObj(prefix='/', command='test', mention='') None`

y-young:
# Description

Fixes #556 

## Type of change

# How Has This Been Tested?

- [x] Parse command with no argument
- [x] Parse command with one argument
- [x] Parse command with more than one arguments
- [x] Print `CommandObj.text`

* Python version: 3.8.1

# Checklist:

### DIFF
--- a/aiogram/dispatcher/filters/builtin.py
+++ b/aiogram/dispatcher/filters/builtin.py
@@ -4,7 +4,7 @@ import typing
 import warnings
 from contextvars import ContextVar
 from dataclasses import dataclass, field
-from typing import Any, Dict, Iterable, Optional, Union
+from typing import Any, Dict, Iterable, List, Optional, Union
 
 from babel.support import LazyProxy
 
@@ -111,6 +111,7 @@ class Command(Filter):
             return False
 
         full_command = text.split()[0]
+        args = text.split()[1:]
         prefix, (command, _, mention) = full_command[0], full_command[1:].partition('@')
 
         if not ignore_mention and mention and (await message.bot.me).username.lower() != mention.lower():
@@ -120,7 +121,7 @@ class Command(Filter):
         if (command.lower() if ignore_case else command) not in commands:
             return False
 
-        return {'command': cls.CommandObj(command=command, prefix=prefix, mention=mention)}
+        return {'command': cls.CommandObj(command=command, prefix=prefix, mention=mention, args=args)}
 
     @dataclass
     class CommandObj:
@@ -137,7 +138,7 @@ class Command(Filter):
         """Mention (if available)"""
         mention: str = None
         """Command argument"""
-        args: str = field(repr=False, default=None)
+        args: List[str] = field(repr=False, default_factory=list)
 
         @property
         def mentioned(self) -> bool:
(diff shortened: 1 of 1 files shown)


aiogram/dispatcher/filters/builtin.py
```diff
@@ -137,7 +138,7 @@ class CommandObj:
         """Mention (if available)"""
         mention: str = None
         """Command argument"""
-        args: str = field(repr=False, default=None)
+        args: List[str] = field(repr=False, default_factory=list)
```
JrooTJunior, line 141: The same as in previous comment.

aiogram/dispatcher/filters/builtin.py
```diff
@@ -111,6 +111,7 @@ async def check_command(cls, message: types.Message, commands, prefixes, ignore_
             return False
 
         full_command = text.split()[0]
+        args = text.split()[1:]
```
JrooTJunior, line 114: Command arguments should not be a list because arguments can be parsed in different ways.
In most of cases it should be string (not list of strings) and this variant is most versatile.

So that's mean this code should be changed to something like this:
```python3
full_command, *args_list = text.split(maxsplit=1)
args = args_list[0] if args_list else None
```

JrooTJunior (changes requested):

JrooTJunior:
Also you can check how it works in `dev-3.x` branch
https://github.com/aiogram/aiogram/blob/dev-3.x/aiogram/dispatcher/filters/command.py#L47-L104

y-young:
@JrooTJunior Could you please review the changes?

aiogram/dispatcher/filters/builtin.py
```diff
@@ -159,7 +160,7 @@ def text(self) -> str:
             if self.mentioned:
                 line += '@' + self.mention
             if self.args:
-                line += ' ' + self.args
+                line += ' ' + ' '.join(self.args)
```
evgfilim1, line 163: `self.args` is a `str`, you forgot to revert this
```suggestion
                line += ' ' + self.args
```

evgfilim1 (changes requested):

y-young:
@JrooTJunior @evgfilim1 Please review the changes, thanks.

evgfilim1 (approved):
Looks good, thanks for PR!

JrooTJunior (approved):

PR #560, opened 2021-04-09
aiogram.types.reply_keyboard: implemented __iadd__, __add__ methods

floordiv:
# Description

Implemented __add__, __iadd__ methods for ReplyKeyboardMarkup entity

## Type of change

# Checklist:

### DIFF
--- a/aiogram/types/reply_keyboard.py
+++ b/aiogram/types/reply_keyboard.py
@@ -90,6 +90,16 @@ class ReplyKeyboardMarkup(base.TelegramObject):
             self.add(button)
         return self
 
+    def __iadd__(self, other):
+        return self.add(other)
+
+    def __add__(self, other):
+        return ReplyKeyboardMarkup(
+            keyboard=self.keyboard + other.keyboard, resize_keyboard=self.resize_keyboard,
+            one_time_keyboard=self.one_time_keyboard, selective=self.selective,
+            row_width=self.row_width
+        )
+
 
 class KeyboardButton(base.TelegramObject):
     """


Olegt0rr:
Mb add similar behavior for inline keyboard?

floordiv:
> Mb add similar behavior for inline keyboard?

Yes, you're right

Olegt0rr:
Mb build class object with `super()` instead it's name?

floordiv:
> Mb build class object with `super()` instead it's name?

I don't think it matters

aiogram/types/inline_keyboard.py
```diff
@@ -82,6 +83,17 @@ def insert(self, button):
...
+    def __iadd__(self, other):
+        return self.add(other)
+
+    def __add__(self, other):
+        return InlineKeyboardMarkup(
+            inline_keyboard=self.inline_keyboard,
+            row_width=self.row_width,
+            conf=self.conf,
+            **self.other_kwargs
+        )
```
Olegt0rr, line 95: `other` is not used here

aiogram/types/inline_keyboard.py
```diff
@@ -82,6 +83,17 @@ def insert(self, button):
             self.add(button)
         return self
 
+    def __iadd__(self, other):
+        return self.add(other)
+
+    def __add__(self, other):
+        return InlineKeyboardMarkup(
```
Olegt0rr, line 90: `self.__class__(...)` - it would be clearer

aiogram/types/reply_keyboard.py
```diff
@@ -90,6 +90,16 @@ def insert(self, button):
             self.add(button)
         return self
 
+    def __iadd__(self, other):
+        return self.add(other)
+
+    def __add__(self, other):
+        return ReplyKeyboardMarkup(
```
Olegt0rr, line 97: `self.__class__(...)` - it would be clearer

Olegt0rr (changes requested):

aiogram/types/reply_keyboard.py
floordiv: I don't think so. Explicit class initializing is better then implicit

Olegt0rr: It's a self class. In this case you know it's behavior without knowing any class names. You shouldn't remember: is it the same class or not.
So it's simple and explicit.

floordiv: done

aiogram/types/inline_keyboard.py
```diff
@@ -82,6 +83,17 @@ def insert(self, button):
             self.add(button)
         return self
 
+    def __iadd__(self, other):
+        return self.add(other)
+
+    def __add__(self, other):
+        return self.__class__(
+            inline_keyboard=self.inline_keyboard + other.keyboard,
```
Olegt0rr, line 91: `other` has no `keyboard` attribute, use `inline_keyboard` instead

Olegt0rr (changes requested):

aiogram/types/inline_keyboard.py
floordiv: thank you, fixed

uwinx:
@floordiv, hey there. thanks for this merge request.

Could you please explain the use-case or motivation for adding magic methods to keyboards markups? Seems a bit "implicit" to me.

floordiv:
> @floordiv, hey there. thanks for this merge request.
> 
> Could you please explain the use-case or motivation for adding magic methods to keyboards' markups? Seems a bit "implicit" to me.

That was an idea of one of the members of Aiogram Telegram chat. It was 3pm so I thought it's pretty good idea.

Use case is pretty easy: using `+=` operator instead of `*KeyboardMarkup.add()`, adding 2 keyboards of the same type (I should add isinstance check) will return new keyboard with buttons from both keyboards:

```
...

keyboard1 = InlineKeyboardMarkup()
keyboard1 += [
    InlineKeyboardButton(text="hello")
]

keyboard2 = InlineKeyboardMarkup()
keyboard2 += [
    InlineKeyboardButton(text="world")
]

...

# this will send keyboard with buttons `hello`, `world`
bot.send_message(chat_id, "What's your choice?", reply_markup=keyboard1 + keyboard2)
```

PR #576, opened 2021-05-04
fix get_full_command for messages with caption

dashedman:
# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A
```python
@dispatcher.message_handler(commands=["command"], commands_ignore_caption=False, content_types=types.ContentType.ANY)
async def message_handler(message: types.Message):
    print( message.get_full_command() )    
```

**Test Configuration**:
* Python version: 3.8

### DIFF
--- a/aiogram/types/message.py
+++ b/aiogram/types/message.py
@@ -194,7 +194,8 @@ class Message(base.TelegramObject):
 
         :return: bool
         """
-        return self.text and self.text.startswith("/")
+        return self.text and self.text.startswith("/") \
+               or self.caption and self.caption.startswith("/")
 
     def get_full_command(self) -> typing.Optional[typing.Tuple[str, str]]:
         """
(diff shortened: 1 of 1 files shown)


aiogram/types/message.py
```diff
@@ -194,7 +194,8 @@ def is_command(self) -> bool:
 
         :return: bool
         """
-        return self.text and self.text.startswith("/")
```
JrooTJunior, line 197: Is more cleaner method is:
```python
test = self.text or self.caption
return text.startswith("/")
```

aiogram/types/message.py
```diff
@@ -203,8 +204,9 @@ def get_full_command(self) -> typing.Optional[typing.Tuple[str, str]]:
         :return: tuple of (command, args)
         """
         if self.is_command():
-            command, *args = self.text.split(maxsplit=1)
-            args = args[-1] if args else ""
+            command, *args = (self.text if self.text
```
JrooTJunior, line 207: The same as previous my comment

JrooTJunior (changes requested):

PR #585, opened 2021-05-11
Added get_url() method for Message object and shifted_id property for Chat object

MasterGroosha:
# Description
Telegram has an ability to generate an URL for a message in group chat or channel.
This PR adds such feature to 3.x branch.
A new method `get_url()` is added with optional `force_private` flag (default to False).
If a group doesn't have a username or `force_private` flag is set, returns URL like `https://t.me/c/shifted_chat_id/message_id`
Otherwise, returns URL like `https://t.me/username/message_id`

This PR also adds `shifted_id` property for Chat object, which is used is private links in `get_url()` method.

## Type of change

# How Has This Been Tested?

With pytest, see "Files changed" section for more info

**Test Configuration**:
* Operating System: Manjaro KDE
* Python version: 3.9

# Checklist:


aiogram/types/message.py
```diff
@@ -1714,6 +1714,26 @@ def copy_to(
...
+        """
+        Returns message URL. Cannot be used in private (one-to-one) chats.
+        If chat has a username, returns URL like https://t.me/username/message_id
+        Otherwise (or if {force_private} flag is set), returns https://t.me/c/shifted_chat_id/message_id
+
+        :param force_private: if set, a private URL is returned even for a public chat
+        :return: string with full message URL
+        """
+        if self.chat.type == "private":
+            raise TypeError("Invalid chat type!")
```
Olegt0rr, line 1727: I suggest to return `None` instead of raising an errror.

Usage with error:
```python
try:
    url = message.get_url()
except TypeError:
    pass
else:
    # action
```

Usage with None:
```python
url = message.get_url()
if url is not None:
    # action
```

aiogram/types/message.py
```diff
@@ -1714,6 +1714,26 @@ def copy_to(
...
+
+        url = "https://t.me/"
+        if not self.chat.username or force_private:
+            url += f"c/{self.chat.shifted_id}/"
```
Olegt0rr, line 1731: f-string + concatenation = bad solution
Since you are using an f-string anyway, just add previous text inside

aiogram/types/message.py
```diff
@@ -1714,6 +1714,26 @@ def copy_to(
...
+        else:
+            url += f"{self.chat.username}/"
+        url += f"{self.message_id}"
```
Olegt0rr, line 1734: f-string + concatenation = bad solution
Since you are using an f-string anyway, just add previous text inside

Olegt0rr (changes requested):

aiogram/types/message.py
MasterGroosha: My approach is based on the current implementation from 2.x
(https://github.com/aiogram/aiogram/blob/dev-2.x/aiogram/types/message.py#L277)

I guess we should discuss this in group chat

aiogram/types/message.py
MasterGroosha: I can replace this with a single template for `.format()`

MasterGroosha: something like `https://t.me/{chat_value}/{message_id}`
Where chat_value is either `/c/` + shifted ID or a username

Olegt0rr: Why `.format()`? 
f-string is better solution. 
Define `chat_value` first and then use f-string

aiogram/types/message.py
MasterGroosha: A note to myself: `if self.chat.type == "private"` should be `self.chat.type in ("private", "group")`

aiogram/types/message.py
```diff
@@ -1723,16 +1723,15 @@ def get_url(self, force_private: bool = False) -> str:
...
 
-        url = "https://t.me/"
         if not self.chat.username or force_private:
-            url += f"c/{self.chat.shifted_id}/"
+            chat_value = f"c/{self.chat.shifted_id}"
         else:
-            url += f"{self.chat.username}/"
-        url += f"{self.message_id}"
-        return url
+            chat_value = f"{self.chat.username}"
```
Olegt0rr, line 1732: Maybe remove redundant f""?
```python
chat_value = self.chat.username
```

aiogram/types/message.py
```diff
@@ -1723,16 +1723,15 @@ def get_url(self, force_private: bool = False) -> str:
         :param force_private: if set, a private URL is returned even for a public chat
         :return: string with full message URL
         """
-        if self.chat.type == "private":
+        if self.chat.type in ("private", "group"):
```
Olegt0rr, line 1726: Why don't you use constants here?
`ChatType.PRIVATE` and `ChatType.GROUP`

MasterGroosha: There is no ChatType class (_yet_)

Olegt0rr (approved):

JrooTJunior (approved):

PR #587, opened 2021-05-17
Сleanup storage

Tishka17:
# Description

No need to store empty data in storage. It can be safely removed
This is done for Redis and Memory storage. It is only implementation details, that is fully compatible with old versions (including real storage internal state)

## Type of change

# How Has This Been Tested?

Added tests for memory storage to check if it is not broken


Tishka17:
**do not merge**:
Check how it working if it nothing to delete in storage. (Probably there could be an error now)

JrooTJunior (approved):

PR #615, opened 2021-07-02
Adding MultiRegexp filter

Forden:
Adding filter to return multiple matches, instead of only first found

# Description

Adding filter to return multiple matches, instead of only first found

## Type of change


uburuntu:
Hello! Could you please add some examples (in docstrings, for example) or tests (or both)? Currently, this filter requires reading its code to understand its purpose, which is not good for libraries used by many people.

PR #669, opened 2021-08-20
Move update type detecting from `Dispatcher` to `Update`

darksidecat:
# Description

Move update type detecting from `Dispatcher` to `Update`

## Type of change

# How Has This Been Tested?

- [x] Run tests locally

**Test Configuration**:
* Operating System: Win10
* Python version: 3.9.2

# Checklist:

### DIFF
--- a/CHANGES/669.misc
+++ b/CHANGES/669.misc
@@ -1,0 +1,1 @@
+move update type detecting to Update
--- a/CHANGES/669.misc
+++ b/CHANGES/669.misc
@@ -1,1 +1,1 @@
-Update type detection is was moved from Dispatcher to Update object
+Moved update type detection from Dispatcher to Update object
--- a/CHANGES/669.misc
+++ b/CHANGES/669.misc
@@ -1,1 +1,1 @@
-move update type detecting to Update
+Update type detection is was moved from Dispatcher to Update object
--- a/aiogram/types/update.py
+++ b/aiogram/types/update.py
@@ -20,6 +20,7 @@ T = TypeVar("T")
 
 
 def lru_cache(func: Callable[..., T]) -> T:
+    """fix lru_cache annotation doesn't work with a property"""
     return functools.lru_cache()(func)  # type: ignore
 
 
--- a/aiogram/types/update.py
+++ b/aiogram/types/update.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
-import functools
-from typing import TYPE_CHECKING, Callable, Optional, TypeVar, cast
+from typing import TYPE_CHECKING, Optional, cast
 
+from ..utils.mypy_hacks import lru_cache
 from .base import TelegramObject
 
 if TYPE_CHECKING:  # pragma: no cover
@@ -15,13 +15,6 @@ if TYPE_CHECKING:  # pragma: no cover
     from .poll_answer import PollAnswer
     from .pre_checkout_query import PreCheckoutQuery
     from .shipping_query import ShippingQuery
-
-T = TypeVar("T")
-
-
-def lru_cache(func: Callable[..., T]) -> T:
-    """fix lru_cache annotation doesn't work with a property"""
-    return functools.lru_cache()(func)  # type: ignore
 
 
 class Update(TelegramObject):
@@ -66,7 +59,7 @@ class Update(TelegramObject):
         return hash((type(self), self.update_id))
 
     @property  # type: ignore
-    @lru_cache
+    @lru_cache()
     def event_type(self) -> str:
         """
         Detect update type
--- a/aiogram/types/update.py
+++ b/aiogram/types/update.py
@@ -59,11 +59,10 @@ class Update(TelegramObject):
         Detect content type
 
         Return update type and content
-        If update type unknown raise UpdateTypeLookupError
+        If update type is unknown, raise UpdateTypeLookupError
 
         :return:
         """
-        event: TelegramObject
         if self.message:
             return "message", self.message
         if self.edited_message:
--- a/aiogram/types/update.py
+++ b/aiogram/types/update.py
@@ -94,7 +94,7 @@ class Update(TelegramObject):
         if self.chat_member:
             return "chat_member"
 
-        raise UpdateTypeLookupError("Unknown update type")
+        raise UpdateTypeLookupError("Update does not contain any known event type.")
 
     @property
     def event(self) -> TelegramObject:
@@ -102,4 +102,4 @@ class Update(TelegramObject):
 
 
 class UpdateTypeLookupError(LookupError):
-    pass
+    """Update does not contain any known event type."""
--- a/aiogram/types/update.py
+++ b/aiogram/types/update.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Optional, Tuple
 
 from .base import TelegramObject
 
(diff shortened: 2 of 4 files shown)


aiogram/types/update.py
```diff
@@ -53,3 +53,62 @@ class Update(TelegramObject):
     """*Optional*. The bot's chat member status was updated in a chat. For private chats, this update is received only when the bot is blocked or unblocked by the user."""
     chat_member: Optional[ChatMemberUpdated] = None
     """*Optional*. A chat member's status was updated in a chat. The bot must be an administrator in the chat and must explicitly specify 'chat_member' in the list of *allowed_updates* to receive these updates."""
+
+    @property
```
evgfilim1, line 57: Shouldn't we add `@lru_cache` also, as in 2.x now?
https://github.com/aiogram/aiogram/blob/3aa40224a23f4cd0d8b824c07ebefceae913be07/aiogram/types/message.py#L108-L110

darksidecat: I got `TypeError: unhashable type: 'Update'` because Pydantic BaseModel is unhashable
Reference https://github.com/samuelcolvin/pydantic/issues/1303

What do you think about possibility of adding hashing to TelegramObject? It seems to me that this is unlikely, since TelegramObject may have unhashable objects, like list[MessageEntity]

I also think that the type of update will not be checked as often as the content_type of message.
Now we have only one check of the update type in thein Dispatcher, I also think this check can be used in middlewares.

darksidecat: last, first - content type position in check chain

```
Run #1
Call times: 1_000_000
Update type last     Cached 0:00:00.221997
Update type last     Ucached 0:00:00.380021
------------------------------------------
Update type first    Cached 0:00:00.221000
Update type first    Ucached 0:00:00.130010
------------------------------------------
Message type last    Cached 0:00:00.290000
Message type last    Ucached 0:00:00.909990
------------------------------------------
Message type first   Cached 0:00:00.296010
Message type first   Ucached 0:00:00.191000

Run #2
Call times: 1_000_000
Update type last     Cached 0:00:00.219978
Update type last     Ucached 0:00:00.382022
--------------------------------------------
Update type first    Cached 0:00:00.220000
Update type first    Ucached 0:00:00.131000
-------------------------------------------
Message type last    Cached 0:00:00.294010
Message type last    Ucached 0:00:00.898993
--------------------------------------------
Message type first   Cached 0:00:00.303007
Message type first   Ucached 0:00:00.190990
```

aiogram/types/update.py
```diff
@@ -53,3 +53,62 @@ class Update(TelegramObject):
...
+    def event(self) -> Tuple[str, TelegramObject]:
```
Olegt0rr, line 58: Can't agree this function naming.

If you call it `property` - it should return property only, not tuple

aiogram/types/update.py
```diff
@@ -53,3 +53,62 @@ class Update(TelegramObject):
...
+        elif self.my_chat_member:
+            update_type = "my_chat_member"
+            event = self.my_chat_member
+        elif self.chat_member:
+            update_type = "chat_member"
+            event = self.chat_member
+        else:
+            raise UpdateTypeLookupError("Unknown update type")
+
+        return update_type, event
```
Olegt0rr, line 110: Many `elif` seems redundant here.
You can you either use `if - return` (look at 2.x), either `for` loop with `getattr` (depends on speed of execution)

aiogram/types/update.py
Olegt0rr: > Shouldn't we add `@lru_cache` also, as in 2.x now?

@evgfilim1 I still can't imagine what you wanna cache on Update detecting. Every update is unique in the most cases. Please provide more info about your suggestion

Olegt0rr (changes requested):

aiogram/types/update.py
darksidecat: I tried to use the following implementation but it was slower than the current one

```python3
 def event(self) -> (str, TelegramObject):
        """
        Return update type and update content
        :return:
        """
        update_fields_names = update.__fields_set__

        update_type = None
        update_event = None

        for field_name in update_fields_names:
            field = getattr(update, field_name)
            if isinstance(field, TelegramObject):
                update_type = field_name
                update_event = field
                break

        if not update_event:
            UpdateTypeLookupError("Unknown update type")
        return update_type, update_event
```

image

aiogram/types/update.py
uwinx: @Olegt0rr, what's wrong with that?

aiogram/types/update.py
uwinx: Just a quick note on this example. It can minified to something like:
```python3
 def event(self) -> (str, TelegramObject):
        update_fields_names = self.__fields_set__

        for field_name in update_fields_names:
            field = getattr(update, field_name)
            if isinstance(field, TelegramObject):
                return field_name, field

        raise UpdateTypeLookupError("Update does not contain any known event type.")
```

darksidecat:
```python3
print(self.__fields_set__)
{'update_id', 'message'}
```

```
Call times: 1000000
Update type last     Loop     0:00:00.556002
Update type last     If       0:00:00.626001
--------------------------------------------------------
Update type first    Loop     0:00:00.610567
Update type first    If       0:00:00.220004
--------------------------------------------------------
Call times: 1000000
Update type last     Loop     0:00:00.617021
Update type last     If       0:00:00.623003
--------------------------------------------------------
Update type first    Loop     0:00:00.613993
Update type first    If       0:00:00.217004
--------------------------------------------------------
Call times: 1000000
Update type last     Loop     0:00:00.627014
Update type last     If       0:00:00.633000
--------------------------------------------------------
Update type first    Loop     0:00:00.614973
Update type first    If       0:00:00.218997
--------------------------------------------------------
```

aiogram/types/update.py
evgfilim1: @darksidecat `functools.cache` is incompatible with Python 3.8, but aiogram 3.0 should support 3.8–3.10

@Olegt0rr After this PR, our users might use this newly introduced field for their own purposes, so caching can help improve speed a bit.

aiogram/types/update.py
```diff
@@ -53,3 +53,46 @@ class Update(TelegramObject):
     """*Optional*. The bot's chat member status was updated in a chat. For private chats, this update is received only when the bot is blocked or unblocked by the user."""
     chat_member: Optional[ChatMemberUpdated] = None
     """*Optional*. A chat member's status was updated in a chat. The bot must be an administrator in the chat and must explicitly specify 'chat_member' in the list of *allowed_updates* to receive these updates."""
+
+    def content(self) -> Tuple[str, TelegramObject]:
+        """
+        Detect content type
+
+        Return update type and content
+        If update type unknown raise UpdateTypeLookupError
```
evgfilim1, line 62:
```suggestion
        If update type is unknown, raise UpdateTypeLookupError
```

aiogram/types/update.py
darksidecat: @evgfilim1 
I could not get a sufficient performance improvement from using the cache, in the case when the field we need is at the beginning of the IF chain, we get a performance degradation.

Perhaps you can run the tests described above and see if it can be improved

```
Call times: 1_000_000
Update type last     Cached   0:00:00.221997
Update type last     Uncached 0:00:00.380021
------------------------------------------
Update type first    Cached   0:00:00.221000
Update type first    Uncached 0:00:00.130010
```

aiogram/types/update.py
```diff
@@ -15,6 +16,13 @@
     from .pre_checkout_query import PreCheckoutQuery
     from .shipping_query import ShippingQuery
 
+T = TypeVar("T")
+
+
+def lru_cache(func: Callable[..., T]) -> T:
```
JrooTJunior, line 22: This hack can be used in other places later, so that's good reason to move this function to utils

CHANGES/669.misc
```diff
@@ -0,0 +1 @@
+move update type detecting to Update
```
JrooTJunior, line 1: Changelog should describe what changes have been made and not what is changing at the moment
```suggestion
Update type detection is was moved from Dispatcher to Update object
```

In this PR you make a movement of the functionality, so in the changelog it should be described as already moved or already added and etc.

JrooTJunior (changes requested):

CHANGES/669.misc
```diff
@@ -1 +1 @@
-move update type detecting to Update
+Update type detection is was moved from Dispatcher to Update object
```
evgfilim1, line 1: Fix typo or mistake, whatever it is.

```suggestion
Update type detection was moved from Dispatcher to Update object
```

I'd also prefer putting the verb first, like in (imo good) commit messages.

```suggestion
Moved update type detection from Dispatcher to Update object
```

evgfilim1 (changes requested):
I think we might need someone who re-checks [~~adwanced~~](https://github.com/aiogram/aiogram/commit/1cdfa3e3c2f11d8b7e0e7c137be415d2ee5bfd3c) English sentences after @JrooTJunior :sweat_smile:

PR #687, opened 2021-09-07
GitHub Action to run pre-commit

cclauss:
# Description

Results: https://github.com/cclauss/aiogram/actions

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Operating System: 
* Python version: 

# Checklist:

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,16 @@
+# https://pre-commit.com
+# This GitHub Action assumes that the repo contains a valid .pre-commit-config.yaml file.
+name: pre-commit
+on: [pull_request, push]
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "3.8"
+      - run: pip install pre-commit poetry
+      - run: pre-commit --version
+      - run: pre-commit install
+      - run: pre-commit run --all-files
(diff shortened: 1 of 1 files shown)


JrooTJunior:
Why do you add new action?
black, isort, flake8, mypy checks is already used at build action directly check on multiple platforms.

I don't see any reasons to add one more action with the same scenario

cclauss:
black, isort, flake8, mypy Do not need to be run on multiple operating systems because they will provide identical results.  So running them once in a quick job provides rapid feedback to contributors who have not yet configured pre-commit on their local machines.  This moves configuration complexity out of GitHub Actions so our non-pre-commit jobs will run faster and in parallel.

JrooTJunior:
It's wrong because new versions of Python has changes in syntax and code should be checked for each supported version

JrooTJunior:
And now after you changed the action with using the matrix build, why we need to provide two separate actions with the same functionality?

JrooTJunior:
So.. As the result..
I dont like `pre-commit` util because in the past it doesn't works correctly with customized black, isort, mypy configs and also it doesn't respect the `dependencies`/`dev-dependencies` section of `pyproject.toml` and can use different version of required `black`, `isort`, `flake8`, `mypy` i won't to use only `pre-commit` to run checks in my projects, i'll recommend you to add step with `pre-commit` to `tests.yml` action as new step without replacion old ones. In this case we have directly call to the linters and then run `pre-commit` with additional check, when something wrong with version or configs we can always see the source of this errors.

PR #688, opened 2021-09-09
custom filter for filters_factory example

Fenicu:
New example of using BoundFilter inheritance

### DIFF
--- a/examples/custom_filter_example.py
+++ b/examples/custom_filter_example.py
@@ -1,4 +1,4 @@
-from typing import Union
+from typing import Any, Union
 from aiogram import Bot, Dispatcher, executor, types
 from aiogram.dispatcher.filters import BoundFilter
 from aiogram.dispatcher.handler import ctx_data
@@ -22,18 +22,16 @@ dp = Dispatcher(bot)
 class UserMiddleware(LifetimeControllerMiddleware):
     skip_patterns = ["error", "update"]
 
-    def __init__(self):
-        super(UserMiddleware, self).__init__()
-
     async def pre_process(
         self,
-        update: Union[types.Message, types.CallbackQuery],
+        obj: Union[types.Message, types.CallbackQuery],
         data: dict,
     ):
-        if update.from_user.id not in fake_db:
-            fake_db[update.from_user.id] = {"ban": False}
+        user = obj.from_user
+        if user.id not in fake_db:
+            fake_db[user.id] = {"ban": False}
 
-        data["user_in_db"] = fake_db[update.from_user.id]
+        data["user_in_db"] = fake_db[user.id]
 
 
 class BanUserFilter(BoundFilter):
@@ -44,14 +42,14 @@ class BanUserFilter(BoundFilter):
     key = "banned"
 
     def __init__(self, banned: bool):
-        if banned is False:
-            raise ValueError("filter banned cannot be False")
+        self.banned = banned
 
-    async def check(self, obj: Union[types.Message, types.CallbackQuery]):
-        data = ctx_data.get()
-        user_in_db = data["user_in_db"]
-
-        return not user_in_db["ban"]
+    async def check(self, obj: Any):
+        data: dict = ctx_data.get()
+        user_in_db = data.get("user_in_db", None)
+        if user_in_db:
+            return self.banned == user_in_db["ban"]
+        return False  # If the user is not in the database, then he cannot be banned
 
 
 class IsEvenIdFilter(BoundFilter):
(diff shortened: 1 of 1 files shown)


Olegt0rr (changes requested):
Wait for my review, please, I'll update it soon

Olegt0rr (changes requested):

examples/custom_filter_example.py
```diff
@@ -70,31 +68,56 @@ async def check(self, obj: Union[types.Message, types.CallbackQuery]):
...
+@dp.message_handler(data_in_db=6)
+async def handler_start2(message: types.Message):
+    await message.answer("There are too many records in the database!")
+
+
+@dp.message_handler(is_even=True)
+async def handler_start3(message: types.Message, is_even: bool):
     if is_even:
-        await message.answer("Congratulations, you id is even!")
+        await message.answer("Congratulations, your id is even!")
```
Olegt0rr, line 118: User `equal` instead of `even` if you're not a Shakespeare :)

examples/custom_filter_example.py
```diff
@@ -0,0 +1,125 @@
...
+async def handler_start2(message: types.Message):
+    await message.answer("There are too many records in the database!")
+
+
+@dp.message_handler(is_even=True)
+async def handler_start3(message: types.Message, is_even: bool):
+    if is_even:
+        await message.answer("Congratulations, your id is even!")
+    else:
+        await message.answer("Congratulations, your id is not even :(")
```
Olegt0rr, line 120: This part of code is unreachable, cause `is_even=True` filter is present.

examples/custom_filter_example.py
```diff
@@ -0,0 +1,125 @@
...
+    2222222222: {"ban": True},
+    3333333333: {"ban": False},
+}
+
+
+bot = Bot(token=API_TOKEN)
+dp = Dispatcher(bot)
+
+
+class UserMiddleware(LifetimeControllerMiddleware):
```
Olegt0rr, line 22: Should we isolate topics: filters / middleware?
Or is a complex example better?

\+ @aiogram/dev-team

Olegt0rr (changes requested):

examples/custom_filter_example.py
Fenicu: I think working with the context can be taken as a separate example

examples/custom_filter_example.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+@dp.message_handler(commands="is_even", is_even=1)
+async def handle_not_even_id(message: types.Message):
+    await message.answer("Sorry, but your id... is not even....")
+
+
+@dp.message_handler(data_in_db=LIMIT_DATABASE_RECORDS)
+async def handle_full_database(message: types.Message, data_in_db: int):
+    await message.answer(f"Too many records in the database.\nLimit {LIMIT_DATABASE_RECORDS}, current {data_in_db}")
```
darksidecat, line 104: The message indicates "Too many records in the database" when we have only 5 records at the limit of 6. Also, this message will be sent regardless of the number of records in database.
data_in_db has a value of 1 which is equal to the number of records up to the limit.

examples/custom_filter_example.py
```diff
@@ -0,0 +1,114 @@
...
+async def handle_even_id(message: types.Message):
+    await message.answer("Congratulations, your id is even!")
+
+
+@dp.message_handler(commands="is_even", is_even=1)
+async def handle_not_even_id(message: types.Message):
+    await message.answer("Sorry, but your id... is not even....")
+
+
+@dp.message_handler(data_in_db=LIMIT_DATABASE_RECORDS)
```
darksidecat, line 102:
```suggestion
@dp.message_handler(commands="data_in_db", data_in_db=LIMIT_DATABASE_RECORDS)
```
need a command to run the handler

examples/custom_filter_example.py
```diff
@@ -0,0 +1,114 @@
...
+async def handle_full_database(message: types.Message, data_in_db: int):
+    await message.answer(f"Too many records in the database.\nLimit {LIMIT_DATABASE_RECORDS}, current {data_in_db}")
+
+
+@dp.message_handler(global_admin=True)
```
darksidecat, line 107:
```suggestion
@dp.message_handler(commands="admin", global_admin=True)
```
need a command to run the handler

examples/custom_filter_example.py
```diff
@@ -0,0 +1,114 @@
...
+
+    async def check(self, obj: Union[types.Message, types.CallbackQuery]):
+        user_id = obj.from_user.id
+        return bool(user_id % self.is_even == 0)
+
+
+class DatabaseIsNotEmptyFilter(BoundFilter):
+    """
+    Check if the bot admins users contains the required number of records
+    """
```
darksidecat, line 64: Inconsistent filter name and docstring.
Also, the filter always returns a positive result.

examples/custom_filter_example.py
darksidecat: maybe also add handler for ```global_admin=False``` ?

examples/custom_filter_example.py
Olegt0rr: Why it can't be any message?

darksidecat: @Olegt0rr 
Now there are several handlers that will react to any message:
`handle_full_database` and `handle_admins`

Therefore, only the first handler will always triggered.

I want the user to be able to run any handler to check how the filters work.

Without this, some of the handlers may be unattainable for execution. (In this case, handle_admins)

PR #719, opened 2021-10-05
Enable Python 3.10 for tests

JrooTJunior:


JrooTJunior:
- `async_lru` is not compatible with 3.10: https://github.com/aio-libs/async-lru/issues/383
- poetry should be updated: https://github.com/python-poetry/poetry/issues/4210

PR #782, opened 2021-12-16
add feature of autoregistering routers

drforse:
# Description

Functionality helping to find routers under directory (including subdirectories, sub-subdirectories etc) insteadof importing all routers by hand.
I find importing all routers to be a pretty boring routine, so I was thinking about some way to help automate the process and came up with solution in this pull request.
Please note, that the library's Router class was changed a bit, if that's a problem, may be I could create separate class, smth like IndexedRouter, but I ain't sure if that's a good idea as it might be more confusing than it seems (for library users, I mean)

## Type of change

# How Has This Been Tested?

- [x] I tested code that is now included in examples/routers_autoload. While testing I've changed some parameters to see different behavior.
- [x] Also there are tests included in test directory

**Test Configuration**:
* Python version: 3.8.8

# Checklist:

### DIFF
--- a/aiogram/dispatcher/router.py
+++ b/aiogram/dispatcher/router.py
@@ -5,6 +5,7 @@ from typing import Any, Dict, Generator, List, Optional, Set, Union
 
 from ..types import TelegramObject
 from ..utils.imports import import_module
+from ..utils.mixins import KeepRefsMixin
 from ..utils.warnings import CodeHasNoEffect
 from .event.bases import REJECTED, UNHANDLED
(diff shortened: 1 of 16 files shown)


JrooTJunior:
Dynamic modules loader is bad way to setup application, so this feature should be discussed internally.

drforse:
> Dynamic modules loader is bad way to setup application, so this feature should be discussed internally.

Oh, can you explain how is it bad though?

JrooTJunior:
This feature is too implicit and will be harmful to developers

PR #804, opened 2022-01-03
Add `protect_content` to webhook responses

bomzheg:
# Description

Webhook responses now does not support protect content

## Type of change

**Test Configuration**:
* Operating System:  Windows 10
* Python version: 3.8

# Checklist:


aiogram/dispatcher/webhook.py
```diff
@@ -448,6 +448,17 @@ def _global_disable_web_page_preview():
...
 
 
+class ProtectContentMixin:
+    def protect_content(self):
+        """
+        Protect content
+
+        :return:
+        """
+        setattr(self, "protect_content", True)
```
Olegt0rr, line 458: Why default value is `True`?

aiogram/dispatcher/webhook.py
```diff
@@ -448,6 +448,17 @@ def _global_disable_web_page_preview():
             return bot.disable_web_page_preview
 
 
+class ProtectContentMixin:
```
Olegt0rr, line 451: Why do we need a mixin? Describe a case

Olegt0rr (changes requested):

aiogram/dispatcher/webhook.py
bomzheg: protect_content is verb describe action of protection of content. this method like https://github.com/aiogram/aiogram/blob/4d2d81138681d730270819579f22b3a0001c43a5/aiogram/dispatcher/webhook.py#L418 and https://github.com/aiogram/aiogram/blob/4d2d81138681d730270819579f22b3a0001c43a5/aiogram/dispatcher/webhook.py#L429 created for modify response by invoke method (is cleaner solution than change field of object)

it is not default value. This method change value of field.

aiogram/dispatcher/webhook.py
bomzheg: for consistency with other mixins like https://github.com/aiogram/aiogram/blob/dev-2.x/aiogram/dispatcher/webhook.py#L417  and https://github.com/aiogram/aiogram/blob/4d2d81138681d730270819579f22b3a0001c43a5/aiogram/dispatcher/webhook.py#L428
Client code can invoke method protect_content on created object of Response for protecting message content

Olegt0rr (approved):

evgfilim1 (approved):

evgfilim1:
After reviewing, I've found that some methods aren't implemented as reply to webhook classes. This might be a subject to another PR.

Missing classes I've discovered:
- [ ] `SendAnimation`
- [ ] `SendPoll`
- [ ] `SendDice`
- [ ] ...

Issue #825, opened 2022-02-01
Failed on aiogram.contrib.fsm_storage.redis.AioRedisAdapterV1.get


hoodwinker:
## Context

* Operating System: MacOS
* Python Version: 3.8.8
* aiogram version: 2.18
* aiohttp version: 3.8.1

## Expected Behavior

Trying to handle messages using template [AlbumMiddleware()
](https://github.com/shevasa/BotCustomDeclaration/blob/57eab3e5ce5c70b4d8f3cf4cd2aa73bb3e18a92f/middlewares/for_albums.py)

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. register AlbumMiddleware()
2. send media group to bot
probably "aiogram_dialog" also matter here

### Failure Logs

```python
2022-01-31 22:37:11,850 - asyncio - ERROR - Task exception was never retrieved
future: <Task finished name='Task-28' coro=<Dispatcher._process_polling_updates() done, defined at /Users/ismirnov/miniconda3/envs/bots/lib/python3.8/site-packages/aiogram/dispatcher/dispatcher.py:407> exception=AttributeError("'NoneType' object has no attribute 'get'")>
Traceback (most recent call last):
  File "/Users/ismirnov/miniconda3/envs/bots/lib/python3.8/site-packages/aiogram/dispatcher/dispatcher.py", line 415, in _process_polling_updates
    for responses in itertools.chain.from_iterable(await self.process_updates(updates, fast)):
  File "/Users/ismirnov/miniconda3/envs/bots/lib/python3.8/site-packages/aiogram/dispatcher/dispatcher.py", line 235, in process_updates
    return await asyncio.gather(*tasks)
  File "/Users/ismirnov/miniconda3/envs/bots/lib/python3.8/site-packages/aiogram/dispatcher/handler.py", line 116, in notify
    response = await handler_obj.handler(*args, **partial_data)
  File "/Users/ismirnov/miniconda3/envs/bots/lib/python3.8/site-packages/aiogram/dispatcher/dispatcher.py", line 256, in process_update
    return await self.message_handlers.notify(update.message)
  File "/Users/ismirnov/miniconda3/envs/bots/lib/python3.8/site-packages/aiogram/dispatcher/handler.py", line 100, in notify
    await self.dispatcher.middleware.trigger(f"pre_process_{self.middleware_key}", args + (data,))
[...]

PR #826, opened 2022-02-01
fix: redis failed with middleware

hoodwinker:
fix for https://github.com/aiogram/aiogram/issues/825


Olegt0rr (changes requested):
Too heavy for highload application, we should init redis earlier (only once)

JrooTJunior:
v2 will no more be supported.

PR #844, opened 2022-02-20
Add support for URI-provided db_name in MongoStorage

altvod:
# Description

Added the option to pass `db_name` as part of the URI.
Say, you have the full URI (as is often done) in your settings file or environment variable. This way you don't have to think of ways to correctly parse `db_name` from the URI yourself in your app code just to pass it as a separate argument to `MongoStorage`. Just add `db_from_uri=True`, and the db_name contained in the URI will override the `'aiogram_fsm'` default.

Also, as minor patch, I added `**self._kwargs` to `AsyncIOMotorClient` initialization for the case when the URI was not provided, and username and password are used to construct it.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Operating System: 
* Python version: 

# Checklist:

### DIFF
--- a/aiogram/contrib/fsm_storage/mongo.py
+++ b/aiogram/contrib/fsm_storage/mongo.py
@@ -21,6 +21,7 @@ STATE = 'aiogram_state'
 DATA = 'aiogram_data'
 BUCKET = 'aiogram_bucket'
 COLLECTIONS = (STATE, DATA, BUCKET)
+DEFAULT_DB_NAME = 'aiogram_fsm'
 
 
 class MongoStorage(BaseStorage):
@@ -42,7 +43,7 @@ class MongoStorage(BaseStorage):
         await dp.storage.wait_closed()
     """
 
-    def __init__(self, host='localhost', port=27017, db_name='aiogram_fsm', uri=None,
+    def __init__(self, host='localhost', port=27017, db_name=DEFAULT_DB_NAME, uri=None,
                  username=None, password=None, index=True, db_from_uri=False, **kwargs):
         self._host = host
         self._port = port
--- a/aiogram/contrib/fsm_storage/mongo.py
+++ b/aiogram/contrib/fsm_storage/mongo.py
@@ -43,14 +43,18 @@ class MongoStorage(BaseStorage):
     """
 
     def __init__(self, host='localhost', port=27017, db_name='aiogram_fsm', uri=None,
-                 username=None, password=None, index=True, **kwargs):
+                 username=None, password=None, index=True, db_from_uri=False, **kwargs):
         self._host = host
         self._port = port
         self._db_name: str = db_name
         self._uri = uri
         self._username = username
         self._password = password
-        self._kwargs = kwargs  # custom client options like SSL configuration, etc.
+        # allows db_name to be provided as part of the URI without passing it also as db_name,
+        # while also allowing compatibility for those relying on the 'aiogram_fsm' default
+        self._db_from_uri = db_from_uri
+        # custom client options like SSL configuration, etc.
+        self._kwargs = kwargs
 
         self._mongo: Optional[AsyncIOMotorClient] = None
         self._db: Optional[AsyncIOMotorDatabase] = None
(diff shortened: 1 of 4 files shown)


Olegt0rr:
Any tests?

altvod:
I will do it, but just haven't had the time for now. I hope, it's ok if this PR just hangs here for a couple more weeks

evgfilim1:
@altvod Any news on this?

altvod:
Sorry for the delay.
I added a test specifically for this parameter and turned on the generic storage tests for mongo

altvod:
So, can you please take a look at it?

.gitignore
```diff
@@ -63,3 +63,6 @@ docs/html
 
 # pynev
 .python-version
+
+# MacOS
+.DS_Store
```
evgfilim1, line 68: :thinking:

aiogram/contrib/fsm_storage/mongo.py
```diff
@@ -70,6 +75,11 @@ async def get_client(self) -> AsyncIOMotorClient:
                     logger = logging.getLogger("aiogram")
                     logger.warning("Run `pip install dnspython==1.16.0` in order to fix ConfigurationError. More information: https://github.com/mongodb/mongo-python-driver/pull/423#issuecomment-528998245")
                 raise e
+
+            if self._db_from_uri:
+                # Since the URI was provided directly, and this flag was specified,
+                # extract db_name from the given URI
+                self._db_name = self._mongo.get_default_database().name
```
evgfilim1, line 82: Maybe it's better to do this in `__init__`?

tests/contrib/fsm_storage/test_storage.py
```diff
@@ -1,8 +1,11 @@
+import asyncio
+
```
evgfilim1, line 2: Unused import

.gitignore
altvod: Strictly speaking, it isn't related to this PR, but still a good thing to have in the .gitignore if editing on a Mac

aiogram/contrib/fsm_storage/mongo.py
altvod: Theoretically speaking - yes. But there is no `self._mongo` in `__init__`. Well, technically there is, but it is `None`.
There's a whole lot of things I would move around in this class, but I'm trying to be as unintrusive as possible here

.gitignore
altvod: Ok, removed it.

tests/contrib/fsm_storage/test_storage.py
altvod: Fixed

aiogram/contrib/fsm_storage/mongo.py
Olegt0rr: Why do we need any flags?
If db is provided directly - use is.
If not - use default db.

evgfilim1: @Olegt0rr because it *might* break someone's code, I guess :thinking:

altvod: Exactly. Maybe someone out there is using it with a URL containing one db_name and explicitly providing a different one. Not very likely, but possible. But I do agree that it would be much cleaner without the flag

Olegt0rr: Deprecations may help us not to collect bad practices for fear of breaking something :D

altvod: So, what do you suggest? Should I break off the compatibility?

Olegt0rr: Database name in the link may be used to make auth only and then user may use any database name available to authorized user. So I don't even understand motivation to make this PR at all. It brings something messing.

JrooTJunior:
v2 will no more be supported.

PR #848, opened 2022-02-25
Update function content_type in message

ProgramUser5353:


Olegt0rr:
Chto za govno?

aiogram/types/message.py
```diff
@@ -110,75 +110,10 @@ class Message(base.TelegramObject):
...
-        if self.voice_chat_started:
-            return ContentType.VOICE_CHAT_STARTED
-        if self.voice_chat_ended:
-            return ContentType.VOICE_CHAT_ENDED
-        if self.voice_chat_participants_invited:
-            return ContentType.VOICE_CHAT_PARTICIPANTS_INVITED
-
+        content_types = ContentTypes.all()
+        for content_type in content_types:
+            if eval(f"self.{content_type}"):
```
evgfilim1, line 115: Don't use eval, it's bad, error-prone and might be vulnerable

evgfilim1 (changes requested):

evgfilim1:
One more question. What is the reason for making this pull request? Why there's no template in your first message describing the changes and their reason?

PR #893, opened 2022-04-22
Added an example of a bot based on a class and asyncio

vortex14:


JrooTJunior:
Please explain the motivation for adding this example?

vortex14:
> Please explain the motivation for adding this example

1. It is not always convenient to use only decorators to describe the bot
2. It is not always necessary to use the native way to get the bot update through the executor.start_polling() method
3. Sometimes the bot needs to be integrated into the API at the same time, for example FastAPI and without using the event loop the executor will block the execution
3. There were no simple examples for class-based code layout.

Olegt0rr:
Use polling from Dispatcher. It's nonblocking.
Anyway, class-based bot is not a solution

examples/echo_class_based_bot_asyncio.py
```diff
@@ -0,0 +1,31 @@
...
+        self.dp = Dispatcher(self.bot)
+
+        self.dp.register_message_handler(self.on_message)
+
+    async def on_message(self, message: types.Message):
+        print(message.get_command())
+        await message.answer(f"Asnwer: {message.text}")
+
+    async def run(self):
+        asyncio.create_task(self.dp.start_polling())
```
evgfilim1, line 23:
```suggestion
        await self.dp.start_polling()
```

examples/echo_class_based_bot_asyncio.py
```diff
@@ -0,0 +1,31 @@
...
+
+    async def run(self):
+        asyncio.create_task(self.dp.start_polling())  
+
+if __name__ == '__main__':
+    bot = AsyncioBot()
+    loop = asyncio.get_event_loop()
+    loop.create_task(bot.run())
+
+    loop.run_forever()
```
evgfilim1, line 30:
```suggestion
    asyncio.run(bot.run())
```

evgfilim1 (changes requested):
I don't understand why you spawn tasks…

vortex14:
> 



> Use polling from Dispatcher. It's nonblocking. Anyway, class-based bot is not a solution

That is, you design all bots through decorators in one file? Are you using classes to build complex applications?

vortex14:
> I don't understand why you spawn tasks…

Ok, You are right for this case, but I uploaded a new example on FastAPI based and how you apply this polling without create task ??

evgfilim1:
> > Use polling from Dispatcher. It's nonblocking. Anyway, class-based bot is not a solution
> 
> That is, you design all bots through decorators in one file? Are you using classes to build complex applications?

Nothing prevents you to use `dp.register_foo_handler(...)` and split your bot into several files. See #522 also.

vortex14:
> Nothing prevents you to use

ok, but this is different way for create application, I and added a few examples for this, which nothing in example directories. 
Do you think this is a few examples are useless (FastApi, class Based, asyncio)?

evgfilim1:
> Do you think this is a few examples are useless (FastApi, class Based, asyncio)?

I think class-based one is useless (why do you even need this? how can it help?) and FastAPI one is confusing (what is `fastapi_utils` and why do you use it when you can achieve the same without installing extra package?). 

I'd also like to hear an opinion from @JrooTJunior and @Olegt0rr.

vortex14:
> > Do you think this is a few examples are useless (FastApi, class Based, asyncio)?
> 
> I think class-based one is useless (why do you even need this? how can it help?) and FastAPI one is confusing (what is `fastapi_utils` and why do you use it when you can achieve the same without installing extra package?).
> 
> I'd also like to hear an opinion from @JrooTJunior and @Olegt0rr.


1. Skeleton class based architecture without decorators 
2. Example using asyncio library 
3. Example integration fastapi with bot

All this all copy/past without additional research (stackoverflow and etc resource)

Main goal save to time and give a finished skeleton for differents design.

I think the more examples there are the better. 

If persoly these example useless your, they may be usefull for others )
Do you use FastAPI? 

fastapi_utils usefull library for decomposing subrouting. 
I can add and base example fastapi, but in real project better use cbv for separates files and routing.

vortex14:
> > > Use polling from Dispatcher. It's nonblocking. Anyway, class-based bot is not a solution
> > 
> > 
> > That is, you design all bots through decorators in one file? Are you using classes to build complex applications?
> 
> Nothing prevents you to use `dp.register_foo_handler(...)` and split your bot into several files. See #522 also.

Why is the example from the discussion not presented as a separate example in the examples folder?

vortex14:
> I agree that #522 should be added to the docs.
> 
> But I'm against wraping an already class-based Bot with another class. It's redundant. This is an antipattern.
> 
> All explained motivation may be solved other way

ok, write please example how organize code for commands and subcommands bot in different space(file, class or etc)?
1. Command on top level
2. Command on sub level 
3. Command on sub/sub level
main command /user -> /add -> /asign role

please micro example for best practice pattern.

evgfilim1:
> please micro example for best practice pattern.

I think this and further future questions not related to this pull request should be moved into discussions

examples/echo_fastapi_based_bot.py
```diff
@@ -0,0 +1,45 @@
...
+
+        self.dp.register_message_handler(self.on_message)
+
+    async def on_message(self, message: types.Message):
+        print(message.get_command())
+        await message.answer(f"Asnwer: {message.text}")
+
+    async def run(self):
+        # await self.dp.start_polling() # will be blocking main :on_start: FastAPI event
+        asyncio.create_task(self.dp.start_polling())
```
evgfilim1, line 25: FastAPI with polling :/
IMO, this doesn't make any sense since you already have a web server running.

examples/echo_fastapi_cbv_based_bot.py
```diff
@@ -0,0 +1,54 @@
...
+
+        self.dp.register_message_handler(self.on_message)
+
+    async def on_message(self, message: types.Message):
+        print(message.get_command())
+        await message.answer(f"Asnwer: {message.text}")
+
+    async def run(self):
+        # await self.dp.start_polling() # will be blocking main :on_start: FastAPI event
+        asyncio.create_task(self.dp.start_polling())
```
evgfilim1, line 26: Same here

vortex14:
> > please micro example for best practice pattern.
> 
> I think this and future questions not related to this pull request should be moved into discussions

https://github.com/aiogram/aiogram/discussions/910

JrooTJunior:
v2 will no more be supported.

PR #903, opened 2022-05-07
Fix escaping for markdown

venticello:
# Description

Fix escaping forbidden characters for markdown. The wrong decorator was used.

Fixes # (issue)

## Type of change


JrooTJunior (changes requested):
Please add test that prove your fix and prevent to break this feature in the future

Also the short description of changes should be added as described above.

JrooTJunior:
@venticello, any news?

JrooTJunior:
Fixed in dev-3.x directly in due to author of this PR did not respond in too long time

PR #915, opened 2022-05-17
make ChatTypeFilter work with inline queries

abrikk:
# Description

I added the ability to work with the ChatTypeFilter filter with inline mode. Also added a new type of chat - sender, also added it to the documentation in the chat types

## Type of change

# How Has This Been Tested?

- [x] Manual Testing

**Test Configuration**:
* Operating System:  Ubuntu 20.04 LTS
* Python version: 3.8.10

# Checklist:

### DIFF
--- a/aiogram/dispatcher/filters/builtin.py
+++ b/aiogram/dispatcher/filters/builtin.py
@@ -732,7 +732,7 @@ class ChatTypeFilter(BoundFilter):
         if isinstance(obj, Message):
             chat_type = obj.chat.type
         elif isinstance(obj, CallbackQuery):
-            chat_type = obj.message.chat.type
+            chat_type = obj.message.chat.type if obj.message else None
         elif isinstance(obj, ChatMemberUpdated):
             chat_type = obj.chat.type
         elif isinstance(obj, InlineQuery):
--- a/aiogram/dispatcher/filters/builtin.py
+++ b/aiogram/dispatcher/filters/builtin.py
@@ -728,18 +728,20 @@ class ChatTypeFilter(BoundFilter):
 
         self.chat_type: typing.Set[str] = set(chat_type)
 
-    async def check(self, obj: Union[Message, CallbackQuery, ChatMemberUpdated]):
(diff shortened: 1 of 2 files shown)


aiogram/dispatcher/filters/builtin.py
```diff
@@ -728,18 +728,20 @@ def __init__(self, chat_type: typing.Container[ChatType]):
...
         if isinstance(obj, Message):
-            obj = obj.chat
+            chat_type = obj.chat.type
         elif isinstance(obj, CallbackQuery):
-            obj = obj.message.chat
+            chat_type = obj.message.chat.type
         elif isinstance(obj, ChatMemberUpdated):
-            obj = obj.chat
+            chat_type = obj.chat.type
+        elif isinstance(obj, InlineQuery):
```
JrooTJunior, line 738:
```python
... and obj.chat_type
```
because `chat_type` is optional field and can be `None`, we don't need additional actions to check the `self.chat_type` contains `None` or not.

aiogram/dispatcher/filters/builtin.py
```diff
@@ -728,18 +728,20 @@ def __init__(self, chat_type: typing.Container[ChatType]):
...
         self.chat_type: typing.Set[str] = set(chat_type)
 
-    async def check(self, obj: Union[Message, CallbackQuery, ChatMemberUpdated]):
+    async def check(self, obj: Union[Message, CallbackQuery, ChatMemberUpdated, InlineQuery]):
         if isinstance(obj, Message):
-            obj = obj.chat
+            chat_type = obj.chat.type
         elif isinstance(obj, CallbackQuery):
-            obj = obj.message.chat
+            chat_type = obj.message.chat.type
```
JrooTJunior, line 735: Callback query can be without `message`, so, we need to add additional condition to prevent to attribute error.

JrooTJunior (changes requested):

aiogram/dispatcher/filters/builtin.py
abrikk: I did a chat type check in a secret chat and it is really not necessary to do a check, since the result of the type will be `None` and there will be no error, if i understood you properly.

PR #948, opened 2022-06-26
Skip running tests when there are no code or test changes

evgfilim1:
Note that I haven't tested this

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,9 +4,17 @@ on:
   push:
     branches:
       - dev-3.x
+    paths:
+      - "aiogram/**"
+      - "tests/**"
+      - ".github/workflows/tests.yml"
   pull_request:
     branches:
       - dev-3.x
+    paths:
+      - "aiogram/**"
+      - "tests/**"
+      - ".github/workflows/tests.yml"
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -4,9 +4,17 @@ on:
   push:
     branches:
       - dev-3.x
+    paths:
+      - "aiogram/**"
+      - "tests/**"
+      - ".github/workflows/tests.yml"
```
JrooTJunior, line 10: `poetry.lock`, `pyproject.toml`, `.flake8`, `mypy.ini`, `codecov.yaml`, `.coveragerc` should also be added, as we need to check the code after updating the dependencies or configs

.github/workflows/tests.yml
```diff
@@ -4,9 +4,17 @@ on:
...
   pull_request:
     branches:
       - dev-3.x
+    paths:
+      - "aiogram/**"
+      - "tests/**"
+      - ".github/workflows/tests.yml"
```
JrooTJunior, line 17: The same as `pull` section

JrooTJunior (changes requested):

JrooTJunior (approved):

PR #955, opened 2022-07-03
Add full support with PyPy

Ad101DP:
# Description

Despite the fact that Aiogram 3 works with pypy in the little I've tested, I don't know to what extent, or if later it will throws some unexpected error, especially knowing that Aiogram 2.x did not run in pypy, so I suggest adding total and verified support with pypy in Aiogram 3, this is essential if want to have the highest speed possible and stand out even more above the other telegram modules as the fastest, since in pypy the speed with respect to the original CPython is approximately 5 times faster, in some cases until 50 times faster, it's crazy!

## Type of change

- New feature (non-breaking change that adds functionality)

# Checklist:


evgfilim1:
You're trying to merge dev-2.x into dev-3.x without any changes. If you want to contribute a code, fork a repository first. If you want to create an issue, go to "Issues" secion.

PR #962, opened 2022-07-12
Split "Context" textarea field inside Issue template into several required input fields

evgfilim1:
Example: https://github.com/evgfilim1/actions-test/issues/7

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yaml
+++ b/.github/ISSUE_TEMPLATE/bug.yaml
@@ -12,16 +12,31 @@ body:
         - label: I have searched in the issue tracker for similar bug reports, including closed ones
           required: true
 
-  - type: textarea
+  - type: markdown
     attributes:
-      label: Context
-      description: |
-        Please provide as much information as possible. This will help us to reproduce the issue and fix it.
       value: |
-        - Operating system: e.g. Ubuntu 20.04.2 LTS
-        - Python Version: e.g. 3.10.1
-        - aiogram version: e.g. 2.21 or 3.0b3
-        - aiohttp version: e.g. 3.8.1
+        ## Context
+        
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug.yaml
```diff
@@ -12,16 +12,31 @@ body:
...
+    attributes:
+      label: Python version
+      placeholder: e.g. 3.10.1
+    validations:
+      required: true
+
+  - type: input
+    attributes:
+      label: aiogram version
+      placeholder: e.g. 2.21 or 3.0b3
```
uwinx, line 39: Maybe add `options` enumeration for major revision? 2.x / 3.x and allow specifying x (the rest of the revision: minor, patch, etc.)

.github/ISSUE_TEMPLATE/bug.yaml
```diff
@@ -12,16 +12,31 @@ body:
...
+    attributes:
+      label: Operating system
+      placeholder: e.g. Ubuntu 20.04.2 LTS
+    validations:
+      required: true
+
+  - type: input
+    attributes:
+      label: Python version
+      placeholder: e.g. 3.10.1
```
uwinx, line 32: Maybe add `options` enumeration for supported versions?

.github/ISSUE_TEMPLATE/bug.yaml
evgfilim1: I don't understand the rationale behind it. A user will anyway write a version inside a text field.

.github/ISSUE_TEMPLATE/bug.yaml
evgfilim1: This may be useful, but again, maintaining a list of all possible `major.minor` versions will be a bit hard.

PR #980, opened 2022-08-15
Refactor: inline keboard row

NiKuma0:
Just removed some bad code...

### DIFF
--- a/aiogram/types/inline_keyboard.py
+++ b/aiogram/types/inline_keyboard.py
@@ -63,7 +63,7 @@ class InlineKeyboardMarkup(base.TelegramObject):
         :return: self
         :rtype: :obj:`types.InlineKeyboardMarkup`
         """
-        self.inline_keyboard.append(buttons or [])
+        self.inline_keyboard.append(list(buttons) or [])
         return self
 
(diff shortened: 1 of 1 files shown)


NiKuma0:
now i think it was not bad code...

aiogram/types/inline_keyboard.py
```diff
@@ -55,16 +55,15 @@ def add(self, *args):
...
         Add row
 
-        :param args:
+        :param buttons:
         :return: self
         :rtype: :obj:`types.InlineKeyboardMarkup`
         """
-        btn_array = [button for button in args]
-        self.inline_keyboard.append(btn_array)
+        self.inline_keyboard.append(list(buttons or []))
```
evgfilim1, line 66: This `buttons or []` is useless, when there are no buttons in `buttons`, it will be an empty tuple, so `list(())` will return `[]`

```suggestion
        self.inline_keyboard.append(list(buttons))
```

evgfilim1 (changes requested):

evgfilim1 (approved):
Loogs good

Olegt0rr (approved):

PR #986, opened 2022-08-23
Refactor: changed the `resolve_event_context` method to human-readable

NiKuma0:
Refactor: changed the `resolve_event_context` method of `UserContextMiddleware` class to human-readable

### DIFF
--- a/aiogram/dispatcher/middlewares/user_context.py
+++ b/aiogram/dispatcher/middlewares/user_context.py
@@ -43,32 +43,21 @@ class UserContextMiddleware(BaseMiddleware):
         """
         Resolve chat and user instance from Update object
         """
-        if event.message:
-            return event.message.chat, event.message.from_user
-        if event.edited_message:
-            return event.edited_message.chat, event.edited_message.from_user
-        if event.channel_post:
-            return event.channel_post.chat, None
-        if event.edited_channel_post:
-            return event.edited_channel_post.chat, None
         if event.inline_query:
-            return None, event.inline_query.from_user
-        if event.chosen_inline_result:
-            return None, event.chosen_inline_result.from_user
-        if event.callback_query:
-            if event.callback_query.message:
+            if event.inlinee_query.message:
                 return event.callback_query.message.chat, event.callback_query.from_user
             return None, event.callback_query.from_user
-        if event.shipping_query:
-            return None, event.shipping_query.from_user
-        if event.pre_checkout_query:
-            return None, event.pre_checkout_query.from_user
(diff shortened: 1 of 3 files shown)


Olegt0rr:
Current solution is better than suggested.

aiogram/dispatcher/middlewares/user_context.py
```diff
@@ -43,32 +43,21 @@ def resolve_event_context(cls, event: Update) -> Tuple[Optional[Chat], Optional[
...
+            'chosen_inline_result', 'callback_query',
+            'shipping_query', 'pre_checkout_query',
+            'poll_answer', 'my_chat_member',
+            'chat_member', 'chat_join_request',
+        )
+        for parametr in parametrs:
+            parametr = getattr(event, parametr, None)
+            if parametr:
+                break
+        return getattr(parametr, 'chat', None), getattr(parametr, 'from_user', None)
```
Olegt0rr, line 63: it's wrong for `poll_answer`

Olegt0rr (changes requested):

aiogram/dispatcher/middlewares/user_context.py
```diff
@@ -43,32 +43,21 @@ def resolve_event_context(cls, event: Update) -> Tuple[Optional[Chat], Optional[
...
-            return event.channel_post.chat, None
-        if event.edited_channel_post:
-            return event.edited_channel_post.chat, None
         if event.inline_query:
-            return None, event.inline_query.from_user
-        if event.chosen_inline_result:
-            return None, event.chosen_inline_result.from_user
-        if event.callback_query:
-            if event.callback_query.message:
+            if event.inlinee_query.message:
```
gabbhack, line 47: typo

gabbhack (changes requested):

NiKuma0:
> Current solution is better than suggested.

Why? 
Okay, I understand that maybe my code will run slower than before (by quite a bit). But it's really the most human readable.

JrooTJunior:
> Okay, I understand that maybe my code will run slower than before (by quite a bit). But it's really the most human readable.

We have already tested such a solution 10 months ago. This solution is slower than current by 4 times and this is due to the use of a cycle with the selection of redundant parameters instead of direct access.

The processing speed might be critical for high-loaded bots

As The Zen of Python says - Explicit is better than implicit.

About readability, i don't thing the for-loop with `gettattr` is more readability instead of plain if-conditions.

---

Also your code doesn't follows project code-style (black)

PR #995, opened 2022-08-29
Prevent to crash polling

JrooTJunior:
# Description

Sometime Telegram restart the server with HTTP 429 error, so that's mean we need to handle it in polling mode

## Type of change

- Bug fix (non-breaking change that fixes an issue)

# Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works as expected

### DIFF
--- a/aiogram/dispatcher/dispatcher.py
+++ b/aiogram/dispatcher/dispatcher.py
@@ -8,7 +8,7 @@ from typing import Any, AsyncGenerator, Dict, List, Optional, Union
 
 from .. import loggers
 from ..client.bot import Bot
-from ..exceptions import TelegramAPIError, TelegramNetworkError, TelegramServerError
+from ..exceptions import TelegramAPIError
 from ..fsm.middleware import FSMContextMiddleware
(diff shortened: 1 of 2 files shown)


aiogram/dispatcher/dispatcher.py
```diff
@@ -183,15 +183,15 @@ async def _listen_updates(
...
+            # Request timeout can be lower than session timeout and that's OK.
             # To prevent false-positive TimeoutError we should wait longer than polling timeout
             kwargs["request_timeout"] = int(bot.session.timeout + polling_timeout)
         while True:
             try:
                 updates = await bot(get_updates, **kwargs)
-            except (TelegramNetworkError, TelegramServerError) as e:
-                # In cases when Telegram Bot API was inaccessible don't need to stop polling process
-                # because some of developers can't make auto-restarting of the script
+            except Exception as e:
```
evgfilim1, line 192: Maybe this would be better? Too broad except clause can be bad, even if we won't get any other error.
```suggestion
            except TelegramAPIError as e:
```

PR #998, opened 2022-09-14
Add type-checking to webhook (parse_update)

artemrootman:
# Description

I tried to configure aiogram with [Sendpulse](https://sendpulse.com/), but I get an error when receiving a webhook: 

```
.../aiogram/dispatcher/webhook.py", line 122, in parse_update
    return types.Update(**data)
TypeError: MetaTelegramObject object argument after ** must be a mapping, not a list
```

I figured out that sendpulse sends a hook like this: `[{...}]`

Some minor changes solved this problem.

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How has this been tested?
`$ curl -X POST https://example.com/webhook -d '[{"from": "me"}]'`
`ok%`

## Test Configuration
- Operating system: Ubuntu 20.04.5 LTS
- Python version: Python 3.8.10
- Aiogram version: 2.22.1

# Checklist:

- [ ] My changes generate no new warnings or errors
- [ ] My changes are compatible with minimum requirements of the project

### DIFF
--- a/aiogram/dispatcher/webhook.py
+++ b/aiogram/dispatcher/webhook.py
@@ -116,6 +116,8 @@ class WebhookRequestHandler(web.View):
         :return: :class:`aiogram.types.Update`
         """
         data = await self.request.json()
+        if isinstance(data, list):
+            data = data[0]
         return types.Update(**data)
 
     async def post(self):


evgfilim1:
I don't think this should be integrated into aiogram. It would be better to write your own wrapper using any web server you like (e.g., aiohttp or fastapi) and use `Dispatcher.process_update`: https://github.com/aiogram/aiogram/blob/96ebee26e4f492615af46068165dfe6b3f7f107f/aiogram/dispatcher/dispatcher.py#L242

PR #1002, opened 2022-09-17
Fixed possible incorrect escaping of strings inside collections.

ENCRYPTEDFOREVER:
# Description
By default if you call `str()` on collection python will escape strings inside of it using single quotes, which will cause

> Can't parse custom emoji identifiers json object

error when calling `getCustomEmojiStickers()`. `json.dumps()` escapes strings by double quotes so no error occurs.

## Type of change
- Bug fix (non-breaking change that fixes an issue)

## Test Configuration
- Operating system: Windows 11  22000.708
- Python version: 3.10.6

- [ ] My changes generate no new warnings or errors
- [ ] My changes are compatible with minimum requirements of the project

### DIFF
--- a/aiogram/bot/api.py
+++ b/aiogram/bot/api.py
@@ -166,7 +166,10 @@ def compose_data(params=None, files=None):
 
     if params:
         for key, value in params.items():
-            data.add_field(key, str(value))
+            if isinstance(value, (list, dict)):
+                data.add_field(key, json.dumps(value))
+            else:
+                data.add_field(key, str(value))
 
     if files:
         for key, f in files.items():


JrooTJunior (changes requested):
Fixed is not in the correct way.

You shouldn't modify the `compose_data` method, you should add the `custom_emoji_ids = prepare_arg(custom_emoji_ids)` statement inside method as the same way as the `results` argument in `answer_inline_query` method.

ENCRYPTEDFOREVER:
> Fixed not in the right way.
> 
> You shouldn't modify the `compose_data` method, you should add the `custom_emoji_ids = prepare_arg(custom_emoji_ids)` statement inside method as the same way as the `results` argument in `answer_inline_query` method.

Fixed. Sorry, i'm not very familiar to aiogram's structure

JrooTJunior (approved):

JrooTJunior:
Thank you for your contribution

PR #1008, opened 2022-09-23
Update parts.py

antaares:
Fixed bug in paginate function. 
# Description

The previous code was not working.
I tried this code and it works fine.

- I have performed a self-review of my own code
- My changes are compatible with minimum requirements of the project

### DIFF
--- a/aiogram/utils/parts.py
+++ b/aiogram/utils/parts.py
@@ -44,17 +44,15 @@ def safe_split_text(text: str, length: int = MAX_MESSAGE_LENGTH, split_separator
     return parts
 
 
-def paginate(data: typing.Iterable, page: int = 0, limit: int = 10) -> typing.Iterable:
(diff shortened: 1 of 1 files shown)


evgfilim1 (changes requested):
This was not a bug. Now you have introduced a magic constant and a breaking change. What is "4" in range? Why cannot I change it to anything I like? Why have you removed "page" argument?

antaares:
Sorry, I wrote another code. I am sorry.

Why do we need a page if we have a limit?

evgfilim1:
> Why do we need a page if we have a limit?

Because `paginate` shouldn't return entire list sliced by `len(part) == limit`. It should return only specific page because it's the only thing you need while paginating.

antaares:
Ehhh, I see now, sorry for taking up your time.

PR #1009, opened 2022-09-24
Update message.py

v9lu:
Updated answer_invoice method.
- Added arguments that were in send_invoice, but were missing in answer_invoice

### DIFF
--- a/aiogram/types/message.py
+++ b/aiogram/types/message.py
@@ -756,9 +756,11 @@ class Message(TelegramObject):
         description: str,
         payload: str,
         provider_token: str,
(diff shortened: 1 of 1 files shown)


v9lu:
При переносе с send_invoice, в аргументах забыл удалить reply_to_message_id, ибо т.к. это answer, реплай все равно будет равен None, поэтому reply_to_message_id должен возвращаться в return как None.

JrooTJunior:
@v9lu, are you alive? You should add the changes description as described in comment from **github-actions** bot.
Otherwise I will not be able to accept your changes.

PR #1032, opened 2022-10-19
Add cleanup to SimpleEventIsolation

andrew000:
# Description

Cleaunp will clear unused locks to prevent memory overflow.

**❗️ MUST BE REVIEWED ❗️**

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How has this been tested?

1. `pytest .\tests\test_fsm\storage\ -vv -s --redis redis://localhost:6379/0`
2. I use these named locks (and cleaunup) in the KusBot project and have already tested it on highload.

## Test Configuration
- Operating system: Windows 10 x64
- Python version: e.g. 3.9

# Checklist:

- [x] I have added tests that prove my fix is effective or that my feature works as expected

### DIFF
--- a/aiogram/fsm/storage/base.py
+++ b/aiogram/fsm/storage/base.py
@@ -1,7 +1,7 @@
 from abc import ABC, abstractmethod
 from contextlib import asynccontextmanager
 from dataclasses import dataclass
-from typing import Any, AsyncGenerator, Dict, Optional, Union
+from typing import Any, AsyncGenerator, Dict, Hashable, Optional, Union
 
 from aiogram import Bot
 from aiogram.fsm.state import State
@@ -91,10 +91,11 @@ class BaseStorage(ABC):
 
 
 class BaseEventIsolation(ABC):
+    _locks: Dict[Hashable, Any]
+
     @abstractmethod
     @asynccontextmanager
     async def lock(self, bot: Bot, key: StorageKey) -> AsyncGenerator[None, None]:
-        self._locks: Dict[StorageKey, Any]
         """
         Isolate events with lock.
         Will be used as context manager
--- a/aiogram/fsm/storage/base.py
+++ b/aiogram/fsm/storage/base.py
@@ -94,6 +94,7 @@ class BaseEventIsolation(ABC):
     @abstractmethod
     @asynccontextmanager
     async def lock(self, bot: Bot, key: StorageKey) -> AsyncGenerator[None, None]:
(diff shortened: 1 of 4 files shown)


JrooTJunior:
@evgfilim1, should not be reviewed until changelog is added.

aiogram/fsm/storage/base.py
```diff
@@ -91,6 +91,8 @@ async def close(self) -> None:  # pragma: no cover
 
 
 class BaseEventIsolation(ABC):
+    _locks: Dict[Hashable, Any]
```
JrooTJunior, line 94: Why this attribute is added to the base class?
This attribute is needed only for in-memory events isolation

aiogram/fsm/storage/memory.py
```diff
@@ -66,5 +66,14 @@ async def lock(self, bot: Bot, key: StorageKey) -> AsyncGenerator[None, None]:
         async with lock:
             yield
 
+        self._cleanup(key)
+
     async def close(self) -> None:
         self._locks.clear()
+
+    def _cleanup(self, key: Hashable) -> None:
+        if self._locks[key]._waiters is None:  # type: ignore[attr-defined]
```
JrooTJunior, line 75: getting access to the protected attribute is a bad decision, but you can inherit from the base class and add the necessary public attribute, for example:

```python
from asyncio import Lock, Future
from typing import Optional, Deque


class _SimpleEventIsolationLock(Lock):
    _waiters: Optional[Deque[Future]] = None

    @property
    def has_waiters(self) -> bool:
        return bool(self._waiters)
```
or implement the dunder method `__bool__` and use lock directly in the conditions

tests/test_fsm/storage/test_isolation.py
```diff
@@ -1,29 +1,67 @@
...
+                asyncio.create_task(self._some_task(isolation, bot, self.random_storage_key(bot)))
+            )
+            await asyncio.sleep(0.01)
+
+        await asyncio.gather(*[task for task in tasks if not task.done()])
+        assert len(isolation._locks) == 0
+
+
+@pytest.mark.parametrize("isolation", [pytest.lazy_fixture("redis_isolation")])
+class TestRedisIsolation:
```
JrooTJunior, line 60: I don't think so the original test should be separated into three, but you can implement new one test that prove your changes, it can alsobe test that you already implement here.

JrooTJunior (changes requested):

PR #1050, opened 2022-10-31
Fix Handler unregister

mamsdeveloper:
Add checking for handler_obj.handler

# Description
At Handler unregister method variable registered was unused so removing handlers by handler function was did`t work. That fix add checking of handler_obj.handler.
 
## Type of change
- Bug fix

# How has this been tested?
- Register handler with dp.register_message_handler(some_func) and then call dp.messages_handler.unregister(some_func)

## Test Configuration
-Python version: 3.9.1

### DIFF
--- a/aiogram/dispatcher/handler.py
+++ b/aiogram/dispatcher/handler.py
@@ -74,11 +74,12 @@ class Handler:
         :param handler: callback
         :return:
         """
+        
         for handler_obj in self.handlers:
-            registered = handler_obj.handler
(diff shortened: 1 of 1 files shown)


evgfilim1 (approved):
Seems legit. Have you tested it?

mamsdeveloper:
Yes, at one of my projects. There are no so many point where it can break

JrooTJunior (approved):

PR #1067, opened 2022-11-14
add sqlite fsm storage

andriystr:

PR #1068, opened 2022-11-14
FSM: SQLiteStorage implemented

nawinds:
# Description

This is a new Storage class after Redis. SQLite does not require any additional installation and is easy-to-use in production

## Type of change

- New feature (non-breaking change that adds functionality)

# How has this been tested?

I've used the new storage class at the bot of my own. Although some of the BaseStorage methods are not implemented yet, I find it useful and I hope somebody would also contribute to the idea.

## Test Configuration
- Operating system: Ubuntu 18, Windows 10
- Python version: 3.7

### DIFF
--- a/aiogram/contrib/fsm_storage/sqlite.py
+++ b/aiogram/contrib/fsm_storage/sqlite.py
@@ -1,0 +1,362 @@
+import typing
+
+from aiogram.dispatcher.storage import BaseStorage
+from sqlalchemy import Column, Integer, String, ForeignKey
+import sqlalchemy as sa
+import sqlalchemy.orm as orm
+from sqlalchemy.orm import Session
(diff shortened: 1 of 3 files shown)


evgfilim1 (changes requested):
I don't think it's a good idea to use SQLAlchemy for such simple task :thinking:
Also see #1067: someone is already trying to implement this.

nawinds:
> I don't think it's a good idea to use SQLAlchemy for such simple task 🤔 Also see #1067: someone is already trying to implement this.

Yeah, I saw that just after opening this pull request... As for SQLAlchemy, for me, it was easier to write code in that way and I believe it's more readable than in pure SQL. But you may be right. I'm really looking for the others' opinion.

JrooTJunior:
v2 will no more be supported.

Issue #946, opened 2022-06-25
Webhook example for 3.x


Masynchin:
**Is your feature request related to a problem? Please describe.**
there is no example with webhook for 3.x version of aiogram.

**Describe the solution you'd like**
Webhook example same as for 2.x version.

JrooTJunior:
We have two examples with webhook:
1. With multibot: https://github.com/aiogram/aiogram/blob/dev-3.x/examples/multibot.py
2. With webapp example: https://github.com/aiogram/aiogram/blob/dev-3.x/examples/web_app/main.py

But it's a good point to have example named like `webhook.py`

Masynchin:
I think this should be added to https://github.com/aiogram/aiogram/projects/3#column-18991685

evgfilim1:
> I think this should be added to https://github.com/aiogram/aiogram/projects/3#column-18991685.

Off-topic here, not related to webhook example

JrooTJunior:
Completed: https://docs.aiogram.dev/en/dev-3.x/dispatcher/webhook.html

PR #1075, opened 2022-12-07
Add example for webhook

tshipenchko:
# Description
I added example for webhook. For #946

## Type of change
- Documentation (typos, code examples or any documentation update)

# How has this been tested?
With ngork webhook.py was tested on my local machine

## Test Configuration
- Operating system: NixOS 22.05 (Quokka) 
-Python version: 3.10.8

# Checklist:
- [ ] I have added tests that prove my fix is effective or that my feature works as expected


examples/webhook.py
```diff
@@ -0,0 +1,82 @@
...
+@router.message(F.text)
+async def echo(message: types.Message):
+    # Regular request, add bot: Bot to handler kwargs
+    # await bot.send_message(message.chat.id, message.text)
+
+    # or reply INTO webhook
+    return SendMessage(chat_id=message.chat.id, text=message.text)
+
+
+@router.startup()
```
tshipenchko, line 36: I'm not sure if it's good to use the same router for the TelegramEventObserver and EventObserver. But this is not so important for this example.

examples/webhook.py
```diff
@@ -0,0 +1,82 @@
...
+
+    dispatcher = Dispatcher()
+    dispatcher["webhook_url"] = WEBHOOK_URL
+    dispatcher.include_router(router)
+
+    app = Application()
+    app["bot"] = bot
+
+    # Skip updates
+    # bot.delete_webhook()
```
tshipenchko, line 69: Is this the only way to skip updates?

JrooTJunior: In normally you shouldn't skip the updates, so you shouldn't use it in this example.

examples/webhook.py
JrooTJunior: I'ts ok.

examples/webhook.py
```diff
@@ -0,0 +1,82 @@
...
+
+from aiogram.webhook.aiohttp_server import SimpleRequestHandler, setup_application
+from aiogram import Dispatcher, Router, Bot, types, F
+from aiogram.methods import SendMessage
+
+
+TELEGRAM_TOKEN = getenv("TELEGRAM_TOKEN")
+
+# Webhook settings
+WEBHOOK_HOST = getenv("WEBHOOK_HOST", "https://your.domain")
```
JrooTJunior, line 15: The most frequently asked question about the webhook example in 2.x is the meaning of each property, so could you please explain that in the example?

examples/webhook.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+@router.shutdown()
+async def on_shutdown(bot: Bot, dispatcher: Dispatcher):
+    logging.warning("Shutting down..")
+
+    # Insert code here to run it before shutdown
+
+    # Remove webhook (not acceptable in some cases)
+    await bot.delete_webhook()
```
JrooTJunior, line 48: In which cases?
I think it's not acceptable in most of cases.

examples/webhook.py
```diff
@@ -0,0 +1,82 @@
...
+
+    # Close DB connection (if used)
+    await dispatcher.storage.close()
```
JrooTJunior, line 51: You dont need to close storage when you use the `setup_application` function because it will be closed when dispatcher shutdown is emited: `https://github.com/aiogram/aiogram/blob/dev-3.x/aiogram/dispatcher/dispatcher.py#L88`

examples/webhook.py
```diff
@@ -0,0 +1,82 @@
...
+    # Close bot session
+    await bot.session.close()
```
JrooTJunior, line 53: This action can be moved into framework, for example into `setup_application` function.

examples/webhook.py
```diff
@@ -0,0 +1,82 @@
...
+
+def main():
+    bot = Bot(token=TELEGRAM_TOKEN, parse_mode="HTML")
+
+    dispatcher = Dispatcher()
+    dispatcher["webhook_url"] = WEBHOOK_URL
+    dispatcher.include_router(router)
+
+    app = Application()
+    app["bot"] = bot
```
JrooTJunior, line 66: Why this action is needed?

JrooTJunior (changes requested):

CHANGES/946.doc
```diff
@@ -0,0 +1 @@
+Add example for webhook
```
JrooTJunior, line 1:
```suggestion
Added example for webhook
```

JrooTJunior:
It would be nice if you explain how to use Webhook in the docs.

examples/webhook.py
tshipenchko: I agree. I moved this comment from dev-2.x
https://github.com/aiogram/aiogram/blob/69c9ecb7e01c25df4052b8d7c576ef87f59943eb/examples/webhook_example.py#L47-L48
I'll delete the comment

tshipenchko:
> It would be nice if you explain how to use Webhook in the docs.

Yes, that would be great. But I don't have time to fully describe it in the documentation in the near future. I can temporarily add small comments briefly explaining how the Webhook works.

examples/webhook.py
tshipenchko: Это избыточно, поэтому я удалил

tshipenchko: This is redundantm, so I removed

examples/webhook.py
tshipenchko: What do you mean? I checked source code of setup_application. There is no way to pass functions to shutdown. Do you mean I should change source code and add it?

tshipenchko:
um lol, why dev-3.x got deleted... 
Okay, anyway code quality was not good enough

JrooTJunior:
Seems like you deleted it by yourself. This branch still active on our side.

tshipenchko:
Yeah I see. Anyway, code quality was not good enough)

PR #1081, opened 2022-12-14
Changed `url` in `TelegramEntityTooLarge`

Desiders:
# Description
Changed `url` in `TelegramEntityTooLarge` to a more appropriate documentation link, because [JSON list of RPC errors](https://core.telegram.org/file/464001173/109f8/k9gTkjfzNHs.87735.json/de2c1dc08a54bde597) and error key `ENTITY_BOUNDS_INVALID` contains link to `https://core.telegram.org/api/entities#entity-length` instead of `https://core.telegram.org/bots/api#sending-files`.

## Type of change
- Attribute used as documentation

### DIFF
--- a/aiogram/exceptions.py
+++ b/aiogram/exceptions.py
@@ -103,7 +103,7 @@ class RestartingTelegram(TelegramServerError):
 
 
 class TelegramEntityTooLarge(TelegramNetworkError):
-    url = "https://core.telegram.org/bots/api#sending-files"
+    url = "https://core.telegram.org/api/entities#entity-length"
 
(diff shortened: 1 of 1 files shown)


JrooTJunior:
You are totally incorrect.

The error TelegramEntityTooLarge raises when you are sending file that bigger than allowed.

Source: https://github.com/aiogram/aiogram/blob/a0dbbcfd5d4bb3a8f35b2b821f279e550f406e13/aiogram/client/session/base.py#L110-L111

But you are changing this link to description of text entities.

PR #1146, opened 2023-03-18
Change `InlineQueryResultType.MPEG` to more correct

Desiders:
# Description

Change type of result in `InlineQueryResult` enum for `InlineQueryResultCachedMpeg4Gif]` and `InlineQueryResultMpeg4Gif` to more correct according to documentation

## Type of change


JrooTJunior:
This part of code should not be changed manually, because we uses codegeneration and this fields will be changed again.


You can try to fix it here: https://github.com/aiogram/butcher
And here: https://github.com/aiogram/aiogram/tree/dev-3.x/.butcher

Or just provide an MRE to check this feature.

JrooTJunior (approved):

PR #1158, opened 2023-04-07
Add missing await in examples

hum4noidx:
# Description

Add missing `await` in multibot example

## Type of change

# Checklist:


tshipenchko:
It is not "missing await". It is called reply into webhook. 
Anyway, aiogram executes request in the background if polling is used // webhook 60 second time out
So, please consider closing pull request.
`message.reply(...)` return instance of `SendMessage(...)`. When you `await` it, it will get `Bot` instance from context and execute the request. If your handler returns subtypes of `TelegramMethod`, it will be sent into webhook or called in the backgroud.

PR #1170, opened 2023-04-22
Removed Text filter

JrooTJunior:
Removed text filter in due to is planned to remove this filter few versions ago.

Use `F.text` instead


andrew000:
Press /F to pay respects 🫡

PR #1173, opened 2023-04-30
parameter secret_token

sheldygg:
# Description

Added X-Telegram-Bot-Api-Secret-Token header check

**Test Configuration**:
* Python version: 3.11

# Checklist:

### DIFF
--- a/aiogram/webhook/aiohttp_server.py
+++ b/aiogram/webhook/aiohttp_server.py
@@ -264,6 +264,9 @@ class TokenBasedRequestHandler(BaseRequestHandler):
         self.bot_settings = bot_settings
         self.bots: Dict[str, Bot] = {}
 
+    def verify_secret(self, telegram_secret_token: str) -> bool:
+        pass
+
     async def close(self) -> None:
         for bot in self.bots.values():
             await bot.session.close()
--- a/aiogram/webhook/aiohttp_server.py
+++ b/aiogram/webhook/aiohttp_server.py
@@ -224,7 +224,7 @@ class SimpleRequestHandler(BaseRequestHandler):
         self.secret_token = secret_token
 
     def verify_secret(self, telegram_secret_token: str) -> bool:
-        return telegram_secret_token == self.secret_token
+        return secrets.compare_digest(telegram_secret_token, self.secret_token)
 
     async def close(self) -> None:
         """
--- a/aiogram/webhook/aiohttp_server.py
+++ b/aiogram/webhook/aiohttp_server.py
@@ -132,7 +132,7 @@ class BaseRequestHandler(ABC):
         pass
 
     @abstractmethod
-    def verify_secret(self, telegram_secret_token: str) -> bool:
+    def verify_secret(self, telegram_secret_token: str, bot: Bot) -> bool:
         pass
 
     async def _background_feed_update(self, bot: Bot, update: Dict[str, Any]) -> None:
@@ -189,7 +189,7 @@ class BaseRequestHandler(ABC):
 
     async def handle(self, request: web.Request) -> web.Response:
         bot = await self.resolve_bot(request)
-        if not self.verify_secret(request.headers.get("X-Telegram-Bot-Api-Secret-Token", "")):
+        if not self.verify_secret(request.headers.get("X-Telegram-Bot-Api-Secret-Token", ""), bot):
             return web.Response(body="Unauthorized", status=401)
         if self.handle_in_background:
             return await self._handle_request_background(bot=bot, request=request)
@@ -221,7 +221,7 @@ class SimpleRequestHandler(BaseRequestHandler):
         self.bot = bot
         self.secret_token = secret_token
 
-    def verify_secret(self, telegram_secret_token: str) -> bool:
+    def verify_secret(self, telegram_secret_token: str, bot: Bot) -> bool:
         if self.secret_token:
             return secrets.compare_digest(telegram_secret_token, self.secret_token)
         return True
@@ -261,7 +261,7 @@ class TokenBasedRequestHandler(BaseRequestHandler):
         self.bot_settings = bot_settings
         self.bots: Dict[str, Bot] = {}
 
-    def verify_secret(self, telegram_secret_token: str) -> bool:
+    def verify_secret(self, telegram_secret_token: str, bot: Bot) -> bool:
         return True
 
     async def close(self) -> None:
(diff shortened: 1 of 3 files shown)


aiogram/webhook/aiohttp_server.py
```diff
@@ -184,6 +186,10 @@ async def _handle_request(self, bot: Bot, request: web.Request) -> web.Response:
         return web.Response(body=self._build_response_writer(bot=bot, result=result))
 
     async def handle(self, request: web.Request) -> web.Response:
+        if self.secret_token:
+            t_token = request.headers.get("X-Telegram-Bot-Api-Secret-Token")
```
JrooTJunior, line 190: What does mean prefix `t_`?

I any short prefixes is used in project it should be conventioned in the contribution guide, but as you see we doesn't have any prefixes convention in the project.

So, please set correct and understandable name for the variable.

JrooTJunior (changes requested):
1. Changes is not covered by tests
2. Is not documented
3. Can be applied only for simple static secret token validation (`SimpleRequestHandler`-like handlers wtih static bot instance)
4. Is totally not compatible with multibot-systems

So, this feature should be discussed before implementation

JrooTJunior:
Any news?

sheldygg:
> 

unban in tg chat. I dont quite understand what I need to do

JrooTJunior:
It is unacceptable to justify Russian terrorism and violation of the territorial integrity of Ukraine, I don't want this to happen in my chats, that's why you can't be unbanned.

sheldygg:
I did not justify, I saw a person who wrote something in the chat, and it was written in his channel that the LPR is Russia. I wanted him banned and threw it into the chat. Your wife did it, but then you came and banned me. I don't support Russians, I'm from Ukraine, Chernihiv region

sheldygg:
t.me/kugona
profile of the person who wrote it

sheldygg:
я не виноват =(

JrooTJunior:
But you send the same message

sheldygg:
> But you send the same message

I wanted everyone to see what he had in his channel.  those who communicated at that time in the chat understood what i mean

JrooTJunior:
Glory to the heroes!

Ok, you are unlocked, but you can also visit the Ukrainian chat.

aiogram/webhook/aiohttp_server.py
```diff
@@ -131,6 +133,10 @@ async def resolve_bot(self, request: web.Request) -> Bot:
         """
         pass
 
+    @abstractmethod
+    def secret_verify(self, secret_token: str) -> bool:
```
JrooTJunior, line 137: please change these words (a verb followed by a noun - means do "something" on "thing")

```suggestion
    def veryfy_secret(self, secret_token: str) -> bool:
```

aiogram/webhook/aiohttp_server.py
```diff
@@ -184,6 +190,8 @@ async def _handle_request(self, bot: Bot, request: web.Request) -> web.Response:
         return web.Response(body=self._build_response_writer(bot=bot, result=result))
 
     async def handle(self, request: web.Request) -> web.Response:
+        if not self.secret_verify(request.headers.get("X-Telegram-Bot-Api-Secret-Token")):
```
JrooTJunior, line 193: I believe this check should be performed after the bot instance is already known

This is required to check the secret for a specific bot instance

aiogram/webhook/aiohttp_server.py
```diff
@@ -90,6 +90,7 @@ def __init__(
         self,
         dispatcher: Dispatcher,
         handle_in_background: bool = False,
+        secret_token: Optional[str] = None,
```
JrooTJunior, line 93: This argument is really needed in base request handler?

Is unsuded in base class but only needed in simple reuqest handler.

JrooTJunior (changes requested):

aiogram/webhook/aiohttp_server.py
```diff
@@ -198,16 +204,27 @@ class SimpleRequestHandler(BaseRequestHandler):
...
-        super().__init__(dispatcher=dispatcher, handle_in_background=handle_in_background, **data)
+        super().__init__(
+            dispatcher=dispatcher,
+            handle_in_background=handle_in_background,
+            **data)
         self.bot = bot
+        self.secret_token = secret_token
+
+    def verify_secret(self, telegram_secret_token: str) -> bool:
+        return telegram_secret_token == self.secret_token
```
Olegt0rr, line 227: It's a bad practice to compare secrets via `==`. Use [compare digest](https://docs.python.org/3/library/secrets.html#secrets.compare_digest)

Olegt0rr (changes requested):

aiogram/webhook/aiohttp_server.py
```diff
@@ -198,16 +204,27 @@ class SimpleRequestHandler(BaseRequestHandler):
     """
 
     def __init__(
-        self, dispatcher: Dispatcher, bot: Bot, handle_in_background: bool = True, **data: Any
+        self, dispatcher: Dispatcher,
```
Olegt0rr, line 207:
```suggestion
        self, 
        dispatcher: Dispatcher,
```

aiogram/webhook/aiohttp_server.py
```diff
@@ -185,6 +189,8 @@ async def _handle_request(self, bot: Bot, request: web.Request) -> web.Response:
 
     async def handle(self, request: web.Request) -> web.Response:
         bot = await self.resolve_bot(request)
+        if not self.verify_secret(request.headers.get("X-Telegram-Bot-Api-Secret-Token")):
+            return web.Response(body="Unauthorized", status=401)
```
Olegt0rr, line 193: So what if I don't wanna verify secrets?

sheldygg: i dont have ideas :(

JrooTJunior: You should return `True` by default from `verify_secret` method

Olegt0rr: I think we should make a single choice of implementation level.
- `self.verify_secret` implemented on base level
- `self.secret_token` implemented on upper level

So, we can push down secret token and check it's not None before trying to get headers and verify something.

sheldygg:
hm...

JrooTJunior (approved):

PR #1176, opened 2023-05-10
Fix markdown spoiler

NiKuma0:
# Description
In [telegram bot api documentation](https://core.telegram.org/bots/api#markdownv2-style) said that need two characters, but there we have only one. I did catch a bug with this, but github issue I didn't find and don't want to create.

## Type of change

- Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

I tested it on my own project, but you can use this code:
```python
...
@router.message()
async def test(m: types.Message):
    return await m.answer(m.md_text)
```

### DIFF
--- a/CHANGES/1176.bugfix.rst
+++ b/CHANGES/1176.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed the markdown spoiler parser.
--- a/aiogram/utils/text_decorations.py
+++ b/aiogram/utils/text_decorations.py
@@ -233,7 +233,7 @@ class MarkdownDecoration(TextDecoration):
         return f"~{value}~"
 
     def spoiler(self, value: str) -> str:
-        return f"|{value}|"
+        return f"||{value}||"
 
     def quote(self, value: str) -> str:
         return re.sub(pattern=self.MARKDOWN_QUOTE_PATTERN, repl=r"\\\1", string=value)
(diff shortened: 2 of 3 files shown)


JrooTJunior (changes requested):
Thanks for you contribution!

Please add short explanation of the changes in `CHANGES/1176.bugfix.rst` file and update tests

JrooTJunior:
Wow, It's huge work, cool

I've been thinking about writing my own asgi for webhook before, this is a promising optimization direction and you confirmed my theories, thanks.

I'n not totally understand why you change pydantic to another library, such kind of changes cannot be performed at the current stage of development, it is planned to publish last beta, then few pre-releases (code-freeze, only bug-fixes and small changes can be applied) and then grand release of 3.0.

Also i have plans to change pydantic to dataclasses for 4.0 version, because it is really slow in some cases and provides too many possibilities instead of having simple serialization to solve the current problem. I like pydantic, but not for all tasks, really.

Another point that you have changed all models, but it shouldn't touched manually, it should be changed in Butcher templates (our docs parser and code-generation platform), because i don't want to support new Bot API features manually, i have implemented Butcher instead.

Also as can i see you change multipart/form-data response to application/json, but this is unacceptable in due to files can't be sended to Telegram via JSON response, you can upload file from response only via multipart/form-data.

Conclusion: i can't apply this pull-request, but i'd like to have own asgi server, if you can rework this pull-request without changing pydantic and change JSON to multipart/form-data response and add possibility to use multiple Bot instances on the server please do it. If you can't do it, i can do it by myself after grand release.

JrooTJunior:
image

> ... and update tests

I see that tests still not updated

Just look at test results: https://github.com/aiogram/aiogram/actions/runs/4951994811/jobs/8858509304?pr=1176

JrooTJunior (approved):

PR #1188, opened 2023-06-04
Fill fsm module `__init__.py` file

dsodx:
# Description

Fill `fsm/__init__.py` to simplify imports.
For example use
```python
from aiogram.fsm import FSMContext, StatesGroup, State
from aiogram.fsm import FSMStrategy, apply_strategy
from aiogram.fsm import FSMContextMiddleware
```
instead of
```python
from aiogram.fsm.context import FSMContext
from aiogram.fsm.state import State, StatesGroup
from aiogram.fsm.strategy import FSMStrategy, apply_strategy
from aiogram.fsm.middleware import FSMContextMiddleware
```
NOTE: You should still use

```python

from aiogram.fsm.storage.redis import RedisStorage
from aiogram.fsm.storage.memory import MemoryStorage
from aiogram.fsm.storage.base import BaseStorage
```
to import storages.

## Type of change

# How Has This Been Tested?

- [x] Test A: `pytest tests`: OK

- [x] Test B: Create `test_b.py` at the root of the project, then fill it with:
```python
from aiogram.fsm import (
    FSMContext,
    State,
    StatesGroup,
    StatesGroupMeta,
    FSMContextMiddleware,
    FSMStrategy,
    apply_strategy,
)
```
and run: OK

**Test Configuration**:
* Operating System: `Windows 11 22H2`
* Python version: `3.11.3`

# Checklist:


dsodx:
Ready for review.

sheldygg:
while RedisStorage was not in init, redis was not a required dependency, but now that you have added it to init, redis has become a required dependency

dsodx:
> redis has become a required dependency

Same with magic_filter.

dsodx:
OK. I'll try to fix it.

dsodx:
Fixed.

sheldygg:
i still think its a bad practice because if i try to import RedisStorage and redis is not installed i will get an error that doesn't explain anything

ImportError: Cannot import name 'RedisStorage' from 'aiogram.fsm' (aiogram\fsm\__init__.py)

you need to somehow explain that he does not have redis installed

sheldygg:
well, or all the same, @JrooTJunior will arrange this option

sheldygg:
I think the best option would be to just delete everything related to redis from __init__.py and leave everything else as you did

JrooTJunior:
I don't think so here is `__init__.py` is needed because in common case (separated by modules/packages) you shouldn't place storage, state definition and context usage at the same point.

Anyway if you remove the redis storage at all from the `__init__.py` that's mean  the redis module is not importable from this package at all.

dsodx:
I removed storages from `__init__.py`.

dsodx:
I find this to be a useful PR since, for example,

```python
from aiogram.fsm.context import FSMContext
```

and 

```python
from aiogram.fsm.state import State, StatesGroup
```

are often used together and my PR simplify it.

dsodx:
@JrooTJunior is there any news?

JrooTJunior:
I don't like this change in due to now the redis module is no more importable (still can be imported but is not exported by `__init__`)

dsodx:
so yes, all storages are not exported into init.

PR #1238, opened 2023-07-31
Reworked InputFile sending

JrooTJunior:
# Description

Reworked InputFile sending

### DIFF
--- a/aiogram/types/input_file.py
+++ b/aiogram/types/input_file.py
@@ -41,12 +41,8 @@ class InputFile(ABC):
         self.chunk_size = chunk_size
 
     @abstractmethod
-    async def read(self, chunk_size: int) -> AsyncGenerator[bytes, None]:  # pragma: no cover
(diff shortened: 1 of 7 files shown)


aiogram/types/input_file.py
```diff
@@ -147,8 +143,9 @@ def __init__(
         self.timeout = timeout
         self.bot = bot
 
-    async def read(self, chunk_size: int) -> AsyncGenerator[bytes, None]:
-        stream = self.bot.session.stream_content(
+    async def read(self, bot: "Bot") -> AsyncGenerator[bytes, None]:
+        bot = self.bot or bot
+        stream = bot.session.stream_content(
```
Olegt0rr, line 148: If we only need the session, can it work with the session only (without the redundant bot instance)?
Or bot may be helpful here for some other reasons?

JrooTJunior: It seems like the argument bot can be changed to session here.

Bot is not needed here now (don't know about future), but by the desigh - in most of cases developer should'dn know about some Session but works with bot instance, so, the preffered way to ask for bot instead of session.

Issue #1242, opened 2023-08-02
TypeError: aiogram.dispatcher.router.Router.emit_startup() got multiple values for keyword argument 'bot'


arynyklas:
## Expected Behavior

Successful startup of Dispatcher.start_polling

## Current Behavior

`TypeError: aiogram.dispatcher.router.Router.emit_startup() got multiple values for keyword argument 'bot'`

### Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. `dispatcher.start_polling(bot)`

### Failure Logs

```
Traceback (most recent call last):
  File "...lib\site-packages\aiogram\dispatcher\dispatcher.py", line 519, in start_polling
    await self.emit_startup(bot=bots[-1], **workflow_data)
TypeError: aiogram.dispatcher.router.Router.emit_startup() got multiple values for keyword argument 'bot'
```

JrooTJunior:
Did you pass the "bot" key into `dispatcher.workflow_data`?

With described scenario in the "Steps to reproduce" section it can't be reproduced, but if you set manually `dispatcher.workflow_data["bot"] = bot` this error can be caused.

Olegt0rr:
It also may looks like a `
```python
dp["bot"] = ...
```

PR #1243, opened 2023-08-02
Fix calling Dispatcher.emit_startup (#1242)

arynyklas:
# Description

#1242 

## Type of change

# Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes

### DIFF
--- a/aiogram/dispatcher/dispatcher.py
+++ b/aiogram/dispatcher/dispatcher.py
@@ -546,7 +546,7 @@ class Dispatcher(Router):
             finally:
                 loggers.dispatcher.info("Polling stopped")
                 try:
-                    await self.emit_shutdown(bot=bots[-1], **workflow_data)
+                    await self.emit_shutdown(bots[-1], **workflow_data)
                 finally:
                     if close_bot_session:
                         await asyncio.gather(*(bot.session.close() for bot in bots))
--- a/aiogram/dispatcher/dispatcher.py
+++ b/aiogram/dispatcher/dispatcher.py
@@ -517,6 +517,7 @@ class Dispatcher(Router):
                 **kwargs,
             }
             await self.emit_startup(bots[-1], **workflow_data)
+            workflow_data.pop("bot")
             loggers.dispatcher.info("Start polling")
             try:
                 tasks: List[asyncio.Task[Any]] = [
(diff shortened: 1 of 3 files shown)


aiogram/dispatcher/dispatcher.py
```diff
@@ -546,7 +547,7 @@ async def start_polling(
             finally:
                 loggers.dispatcher.info("Polling stopped")
                 try:
-                    await self.emit_shutdown(bot=bots[-1], **workflow_data)
+                    await self.emit_shutdown(bots[-1], **workflow_data)
```
JrooTJunior, line 550: The same as in startup callback

JrooTJunior (changes requested):

JrooTJunior:
Also please add the short changes description to the `CHANGES/1242.bugfix.rst` file

aiogram/dispatcher/event/handler.py
```diff
@@ -37,6 +38,8 @@ def _prepare_kwargs(self, kwargs: Dict[str, Any]) -> Dict[str, Any]:
         }
 
     async def call(self, *args: Any, **kwargs: Any) -> Any:
+        if "FSMContextMiddleware.close" in str(self.callback):
```
JrooTJunior, line 41: wtf is going on?

JrooTJunior (approved):

aiogram/dispatcher/event/handler.py
arynyklas: I didn't better solution for this bug.

JrooTJunior: It is very bad solution and doesn't make anything better.
You should not change this place, your issue can be fixed in two lines at place where which you do `.pop`.

Just check that key is present in workflow_data and only then pop it, but all other changes should be rolled back.

JrooTJunior:
I don't see any activity in this PR, so, i'll remake it.

JrooTJunior:
Fixed in #1247

PR #1253, opened 2023-08-08
Added a section on Dependency Injection technology in documentation

wakaree:
# Description

I often noticed that developers ignored the existence of the concept of dependency injection, using global variables or singleton instead. Previously, this topic was not covered in the documentation, which I corrected.

## Type of change

# Checklist:

### DIFF
--- a/docs/dispatcher/dependency_injection.rst
+++ b/docs/dispatcher/dependency_injection.rst
@@ -1,12 +1,12 @@
-######
+####################
 Dependency injection
-######
+####################
 
 Dependency injection is a programming technique that makes a class independent of its dependencies. It achieves that by decoupling the usage of an object from its creation. This helps you to follow `SOLID's <https://en.wikipedia.org/wiki/SOLID>`_ dependency inversion and single responsibility principles.
 
 
 How it works in aiogram
-======
+=======================
 For each update :class:`Dispatcher` passes handling context data. Filters and middleware can also make changes to the context.
 
 To access contextual data you should specify corresponding keyword parameter in handler or filter. For example, to get :class:`FSMContext` we do it like that:
@@ -20,7 +20,7 @@ To access contextual data you should specify corresponding keyword parameter in 
 
 
 Injecting own dependencies
-======
+==========================
 
 Aiogram provides several ways to complement / modify contextual data.
 
--- a/docs/dispatcher/dependency_injection.rst
+++ b/docs/dispatcher/dependency_injection.rst
@@ -1,0 +1,56 @@
+######
+Dependency injection
+######
+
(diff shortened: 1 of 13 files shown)


docs/dispatcher/dependency_injection.rst
```diff
@@ -0,0 +1,56 @@
+######
```
JrooTJunior, line 1: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections

> Section headers ([ref](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#sections)) are created by underlining (and optionally overlining) the section title with a punctuation character, **at least as long as the text**

docs/dispatcher/dependency_injection.rst
```diff
@@ -0,0 +1,56 @@
...
+Dependency injection
```
JrooTJunior, line 2: I don't think so this is a clear dependency injection mechanism, but i don't know another name for this feature

examples/context_addition_from_filter.py
```diff
@@ -0,0 +1,33 @@
...
+
+class HelloFilter(Filter):
+    def __init__(self, name: Optional[str] = None) -> None:
+        self.name = name
+
+    async def __call__(
+        self,
+        message: Message,
+        event_from_user: User
+        # as I said previously, filters also can accept keyword parameters like in handlers
```
JrooTJunior, line 19: "As I said previously" is not clear sentence for this place in due to example can be viewed without reading documentation and this comment is first in this file, you say nothing previously.

examples/context_addition_from_filter.py
```diff
@@ -0,0 +1,33 @@
...
+    ) -> Union[bool, Dict[str, Any]]:
+        if message.text.casefold() == "hello":
+            # returning a dictionary that will update the context data
+            return {"name": event_from_user.mention_html(name=self.name)}
+        return False
+
+
+@router.message(HelloFilter(name=None))
```
JrooTJunior, line 27: name is optional argument, you shouldn't pass any balue to it

JrooTJunior (changes requested):

docs/dispatcher/dependency_injection.rst
wakaree: Fixed

docs/dispatcher/dependency_injection.rst
wakaree: > I don't think so this is a clear dependency injection mechanism
The basic concept of dependency injection is clearly visible in the example of passing a router to the current handler (by "event_router" keyword) - the handler itself depends on it and has access to it. 

But you really can pass anything through the handling context and do it both at application startup and at runtime. Nevertheless, I saw other libraries that also name the concept of passing arguments to the function by parameter names / type hints like "depends".

> but i don't know another name for this feature
Maybe "Handling context"?

JrooTJunior (approved):

PR #1298, opened 2023-09-08
update workflow data

sheldygg:
# Description

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Operating System:
* Python version:

# Checklist:

### DIFF
--- a/CHANGES/1298.bugfix
+++ b/CHANGES/1298.bugfix
@@ -1,0 +1,1 @@
+Update workflow data on start
--- a/aiogram/dispatcher/dispatcher.py
+++ b/aiogram/dispatcher/dispatcher.py
@@ -511,16 +511,11 @@ class Dispatcher(Router):
                         signal.SIGINT, self._signal_stop_polling, signal.SIGINT
                     )
 
-            workflow_data = {
-                "dispatcher": self,
-                "bots": bots,
-                **self.workflow_data,
-                **kwargs,
-            }
-            if "bot" in workflow_data:
-                workflow_data.pop("bot")
+            self.workflow_data.update({"dispatcher": self, "bots": bots, **kwargs})
+            if "bot" in self.workflow_data:
+                self.workflow_data.pop("bot")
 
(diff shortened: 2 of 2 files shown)


Olegt0rr:
@sheldygg what's the issue? It works the same, it seems

sheldygg:
> @sheldygg what's the issue? It works the same, it seems

not the same
when i pass kwargs to `start_polling` method they are not added to workflow_data

MRE
```python
import asyncio

from aiogram import Bot, Dispatcher
from aiogram.filters import CommandStart
from aiogram.types import Message

TOKEN = ""
dp = Dispatcher()


@dp.message(CommandStart())
async def command_start_handler(message: Message, some_val: str) -> None:
    assert dp.workflow_data.get("some_val") is not None
    await message.answer(some_val)


async def main() -> None:
    bot = Bot(TOKEN)
    await dp.start_polling(bot, some_val="ok")


if __name__ == "__main__":
    asyncio.run(main())
```

JrooTJunior:
I don't quite understand why it should be added to `workflow_data`.

You should not access `dispatcher.workflow_data` directly from handlers, as you can always retrieve data from it via handler arguments.

sheldygg:
> I don't quite understand why it should be added to `workflow_data`.
> 
> You should not access `dispatcher.workflow_data` directly from handlers, as you can always retrieve data from it via handler arguments.
> 
> Why `start_polling` should modify `workflow_data`?

usually the parameters i need i pass to `Dispatcher.start_polling` kwargs, then i can get it as handler argument

i my case i use `Dispatcher.feed_update` and i must pass all parameters which i already pass to `Dispatcher.start_polling` otherwise i will get something like that `missing 1 required positional argument: 'some_val'`.

to avoid this, I suggest updating the workflow_data so that it can be used like this
`dp.feed_update(bot=bot, update=update, **dp.workflow_data)`

PR #1316, opened 2023-09-21
fixes #206 - emit warning if command starts with command prefix 

### LINKED ISSUE #206: Emit a warning when command starts with command prefix
## Example code
```python
@dp.message_handler(commands=["/help"])
async def get_help(message):
    await message.reply(HELP_TEXT)
```
## Expected output
```
UserWarning: command "/help" starts with command prefix, so handler will trigger only on "//help" command. Remove leading command prefix to avoid this behavior.
```
## Actual output
None

ShubhamBhut:
# Description

Fixes #206 
If the command  starts with command prefix, it removes one prefix automatically and allows users to use both styles of commands: ['/start', '/help'] and ['start', 'help']

## Type of change

-  [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

It was tested locally with pytest. Existing commands list was modified to test this feature.

**Test Configuration**:
* Operating System: NixOS 23.05
* Python version: 3.11.4

# Checklist:

# Note:
I have made some changes in existing test cases in order to test this feature. I believe all functions are still checked as before.

### DIFF
--- a/aiogram/filters/command.py
+++ b/aiogram/filters/command.py
@@ -78,6 +78,17 @@ class Command(Filter):
                 " or their Iterable"
             )
 
+        prefixes = []
+        for char in prefix:
+            if char not in prefixes:
+                prefixes.append(char)
+
+        operations = {
+            BotCommand: lambda cmd, prefix: cmd.command.lstrip(prefix),
+            re.Pattern: lambda cmd, prefix: re.compile(cmd.pattern.lstrip(prefix)),
+            str: lambda cmd, prefix: cmd.lstrip(prefix),
+        }
+
         items = []
         for command in (*values, *commands):
             if isinstance(command, BotCommand):
@@ -89,6 +100,9 @@ class Command(Filter):
                 )
             if ignore_case and isinstance(command, str):
                 command = command.casefold()
+            for individual_prefix in prefixes:
+                if isinstance(command, (str, BotCommand, re.Pattern)):
+                    command = operations[type(command)](command, individual_prefix)
             items.append(command)
 
         if not items:
--- a/aiogram/filters/command.py
+++ b/aiogram/filters/command.py
@@ -78,17 +78,6 @@ class Command(Filter):
                 " or their Iterable"
             )
 
-        prefixes = []
-        for char in prefix:
-            if char not in prefixes:
-                prefixes.append(char)
-
-        operations = {
-            BotCommand: lambda cmd, prefix: cmd.command.lstrip(prefix),
-            re.Pattern: lambda cmd, prefix: re.compile(cmd.pattern.lstrip(prefix)),
-            str: lambda cmd, prefix: cmd.lstrip(prefix),
-        }
-
         items = []
         for command in (*values, *commands):
             if isinstance(command, BotCommand):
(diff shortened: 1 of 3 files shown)


aiogram/filters/command.py
```diff
@@ -50,7 +50,7 @@ def __init__(
         self,
         *values: CommandPatternType,
         commands: Optional[Union[Sequence[CommandPatternType], CommandPatternType]] = None,
-        prefix: str = "/",
+        prefix: str = "/!",
```
Olegt0rr, line 53: We shouldn't edit default prefix

ShubhamBhut: My bad, I forgot to change back after testing.

aiogram/filters/command.py
```diff
@@ -78,6 +78,17 @@ def __init__(
                 " or their Iterable"
             )
 
+        prefixes = []
```
JrooTJunior, line 81: Why do you need to separate prefixes into list of chars?

`for symbol in "abc"` is the same as `for symbol in ["a", "b", "c"]`, so, you don't need to rebuild it in this way.

Also when you need to convert string into list of chars, recommended to use list("abc") instead of for-loop,
also if chars should be unique you can use set instead of list.

aiogram/filters/command.py
```diff
@@ -89,6 +100,9 @@ def __init__(
                 )
             if ignore_case and isinstance(command, str):
                 command = command.casefold()
+            for individual_prefix in prefixes:
+                if isinstance(command, (str, BotCommand, re.Pattern)):
```
JrooTJunior, line 104: because of previous checks, command here can only be `str`, `BotCommand` or `re.Pattern`, so, you don't need to check it additionally.

BotCommand is also impossible case, because o first condition in this loop.

aiogram/filters/command.py
```diff
@@ -78,6 +78,17 @@ def __init__(
                 " or their Iterable"
             )
 
+        prefixes = []
+        for char in prefix:
+            if char not in prefixes:
+                prefixes.append(char)
+
+        operations = {
+            BotCommand: lambda cmd, prefix: cmd.command.lstrip(prefix),
```
JrooTJunior, line 87: You don't need to do anything with BotComand object because this case is impossible after implemented checks.

JrooTJunior (changes requested):
I'm not sure that this review is even necessary after so many comments have been collected.
Can you please provide additional motivation for this feature? Why is this necessary?

JrooTJunior:
@ShubhamBhut, any news?

ShubhamBhut:
Sincere apologies,

I was occupied elsewhere for some days and it completely fell off my mind. If its alright I will update the changes within 24 hours.

ShubhamBhut:
I couldn't find a way to remove prefix from pattern without modifying it, so I decide to remove prefix in `validate_command` stage/function. I hope its alright.

JrooTJunior:
This is not an answer to my questions above

ShubhamBhut:
> I'm not sure that this review is even necessary after so many comments have been collected. Can you please provide additional motivation for this feature? Why is this necessary?

I don't think its necessary to have this feature, it's just more convenient  for users. Feel free to close it if you find this unnecessary. Apologies if this was a waste of time for you.

> Why do you need to separate prefixes into list of chars?
> 
> `for symbol in "abc"` is the same as `for symbol in ["a", "b", "c"]`, so, you don't need to rebuild it in this way.
> 
> Also when you need to convert string into list of chars, recommended to use list("abc") instead of for-loop, also if chars should be unique you can use set instead of list.

I honestly didn't know a direct `for` loop on string was possible in python

PR #1328, opened 2023-10-04
[2.x] Save `_process_polling_updates` from being garbage collected (fixes #628 #927)

Kylmakalle:
# Description
I've been struggling with the mentioned issue a lot, and my Sentry was blowing up with this issue in lots of my bots.
In each of them, I usually call metrics API inside the handler, which is poorly buffered and sometimes not successful or even slow.

Some of my bots are popular enough, so I assume high update volume and slow handlers result in the deallocation of dangling tasks, which results in:
- Tasks are not completed, i.e. last functions in the handler are not called
- Garbage collector messes up everything, and some handlers are unable to complete `current_handler.reset(ctx_token)`, therefore next alive handler pick-ups Context and can't do anything with it

This PR implements reference saving for `_process_polling_updates` Tasks as described in [Python docs](https://docs.python.org/3/library/asyncio-task.html#creating-tasks) under "Important" section.
It may slightly affect the RAM usage of bots.

Fixes #628, Fixes #927 (issue)

## Type of change

Please delete options that are not relevant.

- [ ] Documentation (typos, code examples or any documentation update)
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Production bot. Errors disappeared immediately.

Sentry
Sentry

**Test Configuration**:
* Operating System: Linux, MacOS
* Python version: 3.10.13

# Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

Issue #927, opened 2022-06-18
Task was destroyed but it is pending


koval01:
## Context

I'm using aiogram to design my bot into telegrams, but as the load increased, I started getting a weird bug.

* Operating System: Linux (Heroku-22)
* Python Version: 3.9.13
* aiogram version: 2.20
* aiohttp version: 3.8.1
* uvloop version (if installed): 0.16.0

## Expected Behavior

I expect there was no such mistake.

## Current Behavior

In parallel with receiving this error, the bot ignores requests.

## Failure Information (for bugs)

Task was destroyed but it is pending!
task: <Task pending name='Task-88390' coro=<Dispatcher._process_polling_updates() done, defined at /app/.heroku/python/lib/python3.9/site-packages/aiogram/dispatcher/dispatcher.py:407> wait_for=<_GatheringFuture pending cb=[<TaskWakeupMethWrapper object at 0x7fb1052d0ac0>()]>>

### Steps to Reproduce

The work of the bot under load

### Failure Logs

```
2022-06-18T22:30:32.381817+00:00 app[worker.1]: aiogram.contrib.middlewares.logging - INFO - Process update [ID:844341375]: [failed] (in 607 ms)
2022-06-18T22:30:32.381955+00:00 app[worker.1]: Exception ignored in: <coroutine object Dispatcher.process_update at 0x7fb103ee80c0>
2022-06-18T22:30:32.381968+00:00 app[worker.1]: Traceback (most recent call last):
2022-06-18T22:30:32.381973+00:00 app[worker.1]:   File "/app/.heroku/python/lib/python3.9/site-packages/aiogram/dispatcher/dispatcher.py", line 256, in process_update
2022-06-18T22:30:32.382784+00:00 app[worker.1]:     return await self.message_handlers.notify(update.message)
2022-06-18T22:30:32.382785+00:00 app[worker.1]:   File "/app/.heroku/python/lib/python3.9/site-packages/aiogram/dispatcher/handler.py", line 126, in notify
2022-06-18T22:30:32.382785+00:00 app[worker.1]:     current_handler.reset(ctx_token)
2022-06-18T22:30:32.382786+00:00 app[worker.1]: ValueError: <Token var=<ContextVar name='current_handler' at 0x7fb10a5aba90> at 0x7fb103e5d9c0> was created in a different Context
2022-06-18T22:30:32.382787+00:00 app[worker.1]: Exception ignored in: <coroutine object Handler.notify at 0x7fb103ee8040>
[...]

Issue #628, opened 2021-07-12
ValueError: <Token var=<ContextVar name='current_handler' at 0x7fe4a298a590> at 0x7fe4a0a69b80> was created in a different Context


lintyflent:
## Context
 * Manjaro Linux Gnome Testing branch 
 * Python 3.9.6
 * aiogram 2.11.2
 * aiohttp 3.7.4.post0  

I use Aiomysql in conjunction with Aiogram. Experimentally, I found out that the await con.commit() line causes the error shown at the end. But why? But how is this error related to aiogram and why is there no mention of project files?|
P.s: If I noticed correctly, the error occurs only when UPDATE and possibly INSERT.

```
async def __query(self, *args, **kwargs):
    con = await aiomysql.connect(
            host=self.__db_host,
            user=self.__db_login,
            password=self.__db_pass,
            db=self.__db_name,
    )
    cur = await con.cursor()
    await cur.execute(*args, **kwargs)
    res = await cur.fetchall()
    await con.commit()
    await cur.close()
```
OR
```
...
    async def __query(self, *args, **kwargs):
        pool = await aiomysql.create_pool(
            host=self.__db_host,
            port=self.__port,
            user=self.__db_login,
            password=self.__db_pass,
            db=self.__db_name
        )
        res = None
        async with pool.acquire() as conn:
            async with conn.cursor() as cur:
                await cur.execute(*args, **kwargs)
                res = await cur.fetchall()
                await conn.commit()
        pool.close()
        await pool.wait_closed()
        return res
 ...
    async def execute(self, *args, **kwargs):
        return await self.__query(*args, **kwargs)
...
    async def func(self, args...):
        ...
        await db.execute(f'UPDATE table_name SET column_name = {new_data} WHERE user_id = "{self.user_id}"')
        ...
```

### Steps to Reproduce

Any database query via aiomysql. On commit.

### Failure Logs

The error that occurs:

```
Traceback (most recent call last):
  File "/home/user/.local/lib/python3.9/site-packages/aiogram/dispatcher/handler.py", line 117, in notify
    response = await handler_obj.handler(*args, **partial_data)
GeneratorExit

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/user/.local/lib/python3.9/site-packages/aiogram/dispatcher/dispatcher.py", line 273, in process_update
    return await self.callback_query_handlers.notify(update.callback_query)
[...]

PR #1331, opened 2023-10-07
Prevent update handling task pointers from being garbage collected, backport of #1328

Kylmakalle:
# Description

This is a backport of #1328 to mitigate possible issues like #628, #927. As in #1328, this PR preserves asyncio `Task` pointers until the task is completed to ensure `Task` will not be garbage collected and disappear. See recommendations in [Python docs](https://docs.python.org/3/library/asyncio-task.html#creating-tasks).

> **Important**: Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn’t referenced elsewhere may get garbage collected at any time, even before it’s done. For reliable “fire-and-forget” background tasks, gather them in a collection

## Type of change

# How Has This Been Tested?

Honestly, I don't have a v3-based bot with workload I have in v2. But want to admit once again, it's a pretty rare case and we're handling it as the docs says.

**Test Configuration**:
* Operating System: Macos, Linux
* Python version: 3.11.1

### DIFF
--- a/CHANGES/1331.misc
+++ b/CHANGES/1331.misc
@@ -1,0 +1,1 @@
+Prevent update handling task pointers from being garbage collected, backport from 2.x
--- a/aiogram/dispatcher/dispatcher.py
+++ b/aiogram/dispatcher/dispatcher.py
@@ -6,7 +6,7 @@ import signal
 import warnings
 from asyncio import CancelledError, Event, Future, Lock
 from contextlib import suppress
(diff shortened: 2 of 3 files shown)


Kylmakalle:
I've not seen any evidence of "Task was destroyed but it is pending" for v3 yet, but I assume it's due to a lower adoption rate than v2.

JrooTJunior (approved):

PR #1344, opened 2023-10-20
Add new FSM strategy CHAT_TOPIC strategy.py

OnaZeroN:
# Description

The new FSM strategy, which sets the state for the entire topic in the chat, also works in private messages and regular groups without topics.

aiogram did not have an FSM strategy that would set state only for a specific topic in the chat. This strategy is quite reasonable and it has many applications. 
The work of state in private messages remains unchanged
The work of state in groups / super groups without topics, state will be established for all groups
The work of state in groups /super groups with topics, State will be established only for one topic, other topics will not have a state until it is installed, state can be set separately for each topic, and state_data will be separate in all topics where state is installed.

## Type of change

- New feature (non-breaking change which adds functionality)
- This change requires a documentation update

# How Has This Been Tested?

- Run tests/test_fsm/test_strategy.py
- You can run bot with FSMStrategy.CHAT_TOPIC, and make sure that the strategy is working correctly:

   - Check the work of state in private messages
   - Check the work of state in groups/super groups without topics, State should be set for the whole group, since the topic value will be None
   - Check the work of state in groups/super groups with topics, State should be installed only for one topic, other topics will not have a state until it is installed, state can be set separately for each topic, and state_data for should be separate in all topics where state is installed.

Test Configuration:
* Operating System: windows 11
* Python version: 3.11

# Checklist:

### DIFF
--- a/aiogram/fsm/middleware.py
+++ b/aiogram/fsm/middleware.py
@@ -91,7 +91,7 @@ class FSMContextMiddleware(BaseMiddleware):
         self,
         bot: Bot,
         chat_id: int,
-        user_id: int,
+        user_id: Optional[int] = None,
         thread_id: Optional[int] = None,
         destiny: str = DEFAULT_DESTINY,
     ) -> FSMContext:
--- a/aiogram/fsm/middleware.py
+++ b/aiogram/fsm/middleware.py
@@ -91,7 +91,7 @@ class FSMContextMiddleware(BaseMiddleware):
         self,
         bot: Bot,
         chat_id: int,
-        user_id: Optional[int] = None,
+        user_id: int,
         thread_id: Optional[int] = None,
         destiny: str = DEFAULT_DESTINY,
     ) -> FSMContext:
--- a/aiogram/fsm/storage/base.py
+++ b/aiogram/fsm/storage/base.py
@@ -14,7 +14,7 @@ DEFAULT_DESTINY = "default"
 class StorageKey:
     bot_id: int
     chat_id: int
-    user_id:  int
+    user_id: int
     thread_id: Optional[int] = None
     destiny: str = DEFAULT_DESTINY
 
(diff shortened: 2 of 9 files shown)


aiogram/fsm/strategy.py
```diff
@@ -7,18 +7,20 @@ class FSMStrategy(Enum):
...
-) -> Tuple[int, int, Optional[int]]:
+) -> Tuple[int, Optional[int], Optional[int]]:
     if strategy == FSMStrategy.CHAT:
         return chat_id, chat_id, None
     if strategy == FSMStrategy.GLOBAL_USER:
         return user_id, user_id, None
     if strategy == FSMStrategy.USER_IN_TOPIC:
         return chat_id, user_id, thread_id
+    if strategy == FSMStrategy.CHAT_TOPIC:
+        return chat_id, None, thread_id
```
JrooTJunior, line 25: You should use `chat_id, chat_id, thread_id` instead of `chat_id, None, thread_id`.
Your strategy is very similar to `FSMStrategy.CHAT`.

aiogram/fsm/middleware.py
```diff
@@ -91,7 +91,7 @@ def get_context(
         self,
         bot: Bot,
         chat_id: int,
-        user_id: int,
+        user_id: Optional[int] = None,
```
JrooTJunior, line 94: In case when you use chat_id as user_id in the strategy you don't need to change type-hint here.

aiogram/fsm/storage/base.py
```diff
@@ -14,7 +14,7 @@
 class StorageKey:
     bot_id: int
     chat_id: int
-    user_id: int
+    user_id: Optional[int] = None
```
JrooTJunior, line 17: The same as in previous comment.

JrooTJunior (changes requested):

OnaZeroN:
@JrooTJunior I didn't really understand the error that came out during the tests

JrooTJunior:
> @JrooTJunior I didn't really understand the error that came out during the tests

Is coming from the code-style check.
PEP-8 blank lines: https://peps.python.org/pep-0008/#blank-lines

JrooTJunior (approved):

PR #1370, opened 2023-11-20
Fix CallbackData without default Optional

### LINKED ISSUE #1368: Can`t unpack None value in CallbackData
### Checklist

- [X] I am sure the error is coming from aiogram code
- [X] I have searched in the issue tracker for similar bug reports, including closed ones

### Operating system

Windows10

### Python version

python3.11.6

### aiogram version

aiogram3.1.1

### Expected behavior

chat_id=123 thread_id=None

### Current behavior

Exception

### Steps to reproduce

See code exemple

### Code example

```python3
from aiogram.filters.callback_data import CallbackData

    class TgData(CallbackData, prefix="tg"):
        chat_id: int
        thread_id: int | None


    TgData.unpack('tg:123:456') # ok
    TgData.unpack('tg:123:')  # can`t unpack thread id!
```

### Logs

```sh
pydantic_core._pydantic_core.ValidationError: 1 validation error for TgData thread_id
    Input should be a valid integer, unable to parse string as an integer [type=int_parsing, input_value='', input_type=str]
    For further information visit https://errors.pydantic.dev/2.3/v/int_parsing
```

### Additional information

_No response_

Olegt0rr:
# Description

Fixes #1368

Method `.is_required()` returns `True` on `Optional[int]` without default value, because it's still requires either `int` or `None` value.
On the other hand, when we parse a value from a zero-length string - it cannot be converted to `None` or an `int`, so pyantic returns an error about it.

The solution to this problem is to adapt the field type check to such a case.

### DIFF
--- a/aiogram/filters/callback_data.py
+++ b/aiogram/filters/callback_data.py
@@ -120,9 +120,8 @@ class CallbackData(BaseModel):
             raise ValueError(f"Bad prefix ({prefix!r} != {cls.__prefix__!r})")
         payload = {}
         for k, v in zip(names, parts):  # type: str, Optional[str]
-            if field := cls.model_fields.get(k):
-                if v == "" and not field.is_required():
-                    v = None
+            if v == "":
+                v = None
             payload[k] = v
         return cls(**payload)
 
--- a/aiogram/filters/callback_data.py
+++ b/aiogram/filters/callback_data.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import typing
 from decimal import Decimal
 from enum import Enum
 from fractions import Fraction
@@ -18,6 +19,7 @@ from uuid import UUID
 
 from magic_filter import MagicFilter
 from pydantic import BaseModel
+from pydantic.fields import FieldInfo
 
 from aiogram.filters.base import Filter
 from aiogram.types import CallbackQuery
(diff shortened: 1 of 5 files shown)


aiogram/filters/callback_data.py
```diff
@@ -120,9 +120,8 @@ def unpack(cls: Type[T], value: str) -> T:
             raise ValueError(f"Bad prefix ({prefix!r} != {cls.__prefix__!r})")
         payload = {}
         for k, v in zip(names, parts):  # type: str, Optional[str]
-            if field := cls.model_fields.get(k):
-                if v == "" and not field.is_required():
-                    v = None
+            if v == "":
```
JrooTJunior, line 123: This solution breaks case when field is required with any type (`demo: str` typehint) and value is empty (`demo == ""`)


You should check that typehint of the field is Union and has None argument instead of checking `field.is_required()`

```python
typing.get_origin(field.annotation) is types.UnionType 
and type(None) in typing.get_args(field.annotation)
```

But this solution don't wor in Python <=3.9 because `types.UnionType` is introduced [in the 3.10 version](https://docs.python.org/3/library/types.html#types.UnionType).

We should research alternative universal solution for all currently supported versions of the Python.

JrooTJunior (changes requested):

JrooTJunior (approved):

PR #1398, opened 2024-01-27
Add KeyboardButtonPollTypeType enum

JrooTJunior:
A new enum `KeyboardButtonPollTypeType` for `KeyboardButtonPollTypeType.type` field has bed added.


Donald996:
It should be [KeyboardButtonPollType](https://core.telegram.org/bots/api#keyboardbuttonpolltype),The name seems to be wrong

JrooTJunior:
Object from the Bot API is named as `KeyboardButtonPollType`, but this pull request adds an enumeration of known poll types that can be passed into KeyboardButtonPollType.type field, so, the name of this enum is combination of object type name and field name - `KeyboardButtonPollTypeType`.

Names of the object should **not** be overlapped in the framework codebase.

PR #1399, opened 2024-01-27
Update KeyboardBuilder utility, fixed type-hints for button method, a…

JrooTJunior:
Update KeyboardBuilder utility, fixed type-hints for button method, adjusted limits of the different markup types to real world values.

### DIFF
--- a/aiogram/utils/keyboard.py
+++ b/aiogram/utils/keyboard.py
@@ -106,7 +106,7 @@ class KeyboardBuilder(Generic[ButtonType], ABC):
                 f"not type {type(row).__name__}"
             )
         if len(row) > self.max_width:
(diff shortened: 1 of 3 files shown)


aiogram/utils/keyboard.py
```diff
@@ -138,7 +140,7 @@ def _validate_size(self, size: Any) -> int:
         """
         if not isinstance(size, int):
             raise ValueError("Only int sizes are allowed")
-        if size not in range(MIN_WIDTH, MAX_WIDTH + 1):
+        if size not in range(self.min_width, self.max_width + 1):
             raise ValueError(f"Row size {size} are not allowed")
```
mahenzon, line 144:
```suggestion
            raise ValueError(f"Row size {size} is not allowed, range: [{self.min_width}, {self.max_width}]")
```

mahenzon (approved):

PR #1418, opened 2024-02-17
fix prepare Default

RootShinobi:
# Description

fix TelegramObject from DefaultBotProperties is not JSON serializable 

## Type of change

# Checklist:

### DIFF
--- a/CHANGES/1418.bugfix.rst
+++ b/CHANGES/1418.bugfix.rst
@@ -1,0 +1,1 @@
+fix TelegramObject from DefaultBotProperties is not JSON serializable
--- a/aiogram/client/session/base.py
+++ b/aiogram/client/session/base.py
@@ -38,7 +38,7 @@ from aiogram.exceptions import (
 
 from ...methods import Response, TelegramMethod
 from ...methods.base import TelegramType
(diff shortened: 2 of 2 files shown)


CHANGES/1418.bugfix.rst
```diff
@@ -0,0 +1 @@
+fix TelegramObject from DefaultBotProperties is not JSON serializable
```
JrooTJunior, line 1: Should starts with capitalized word.
Also this line can be not understandable for end-users.

aiogram/client/session/base.py
```diff
@@ -193,6 +193,8 @@ def prepare_value(
             return value
         if isinstance(value, Default):
             default_value = bot.default[value.name]
+            if isinstance(default_value, TelegramObject):
+                default_value = default_value.model_dump(warnings=False)
```
JrooTJunior, line 197: This line should be covered by tests.

More elegant way that is move this condition into one level with other conditions at this function instead of adding one more level.

JrooTJunior (changes requested):
This PR can't be merged in current state, i have already fixed this issue in another PR.

PR #1434, opened 2024-03-06
Added MongoStorage for FSM

Rishat-F:
# Description

1. Added MongoStorage for FSM
2. Added tests for it
3. Docs updated
4. Some code was restructured

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

1. New tests added
2. Added few additional checks in existing tests
3. Code checked with black, isort, mypy

**Test Configuration**:
* Operating System: Ubuntu 20.04 LTS
* Python version: 3.10

# Checklist:


aiogram/fsm/storage/base.py
```diff
@@ -19,6 +19,68 @@ class StorageKey:
...
+            parts.append(str(key.bot_id))
+        parts.append(str(key.chat_id))
+        if key.thread_id:
+            parts.append(str(key.thread_id))
+        parts.append(str(key.user_id))
+        if self.with_destiny:
+            parts.append(key.destiny)
+        elif key.destiny != DEFAULT_DESTINY:
+            raise ValueError(
+                "Default key builder is not configured to use key destiny other the default.\n"
```
JrooTJunior, line 76:
```suggestion
                "Default key builder is not configured to use key destiny other than the default.\n"
```

aiogram/fsm/storage/mongo.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def resolve_state(self, value: StateType) -> Optional[str]:
+        if value is None:
+            return None
+        elif isinstance(value, State):
+            return value.state
+        return str(value)
+
+    async def set_state(self, key: StorageKey, state: StateType = None) -> None:
+        document_id = self._key_builder.build(key, "state")
```
JrooTJunior, line 72: What about using single object key instead of creating multiple documents for the same user?

Just use:
```json5
{
    _id: "fsm:bot_id:user_id:chat_id:destiny", 
    state: state,
    data: {...}
}
```

instead of 
```json5
{
    _id: "fsm:bot_id:user_id:chat_id:destiny:state", 
    state: state,
}
```
and
```json5
{
    _id: "fsm:bot_id:user_id:chat_id:destiny:data",
    data: {...}
}
```

aiogram/fsm/storage/mongo.py
```diff
@@ -0,0 +1,120 @@
...
+                upsert=True,
+            )
+
+    async def get_data(self, key: StorageKey) -> Dict[str, Any]:
+        document_id = self._key_builder.build(key, "data")
+        document = await self._data_collection.find_one({"_id": document_id}, {"_id": 0})
+        if not document:
+            return {}
+        else:
+            return document  # type: ignore
```
JrooTJunior, line 107: Why type ignored?

JrooTJunior, line 107: This statement can be simplified
```suggestion
        if not document:
            return {}
        return document
```

aiogram/fsm/storage/mongo.py
```diff
@@ -0,0 +1,120 @@
...
+        update_result = await self._data_collection.find_one_and_update(
+            {"_id": document_id},
+            {"$set": data},
+            upsert=True,
+            return_document=True,
+            projection={"_id": 0},
+        )
+        if not update_result:
+            await self._data_collection.delete_one({"_id": document_id})
+        return update_result  # type: ignore
```
JrooTJunior, line 120: Why ignore type?

JrooTJunior (changes requested):

aiogram/fsm/storage/mongo.py
Rishat-F: Mypy check throws error:
error: Incompatible return value type (got "Mapping[str, Any]", expected "Dict[str, Any]")

aiogram/fsm/storage/mongo.py
Rishat-F: Sounds reasonable. I will rewrite it within a couple of days. And we will see how it will look like.

Rishat-F: I have thought about this proposal more carefully and came to the conclusion that it is better not to combine state and data into one collection. Due to the following reasons:
1. Most likely we don’t want to store documents with empty "state" key and empty "data" key at the same time (with only "_id" key), because it would be a waste of memory. If so, then there would be more code, more logic, more queries to define if document become empty and remove it.
2. It seems like there no way to update data with one query if it is stored as nested object. So there also needed more code and more queries to do it.
So the main advantage of using two separated collections for state and data over using one union collection is more simple code and fewer queries to MongoDB.

aiogram/fsm/storage/base.py
```diff
@@ -19,6 +19,68 @@ class StorageKey:
     destiny: str = DEFAULT_DESTINY
 
 
+class KeyBuilder(ABC):
+    """Base class for key builder."""
+
+    @abstractmethod
+    def build(self, key: StorageKey, part: Literal["data", "state", "lock"]) -> str:
+        """
+        This method should be implemented in subclasses
```
Olegt0rr, line 28: Redundant comment, `@abstractmethod` already meant this.
Use imperative method description only.

aiogram/fsm/storage/base.py
```diff
@@ -19,6 +19,68 @@ class StorageKey:
...
+
+        :param key: contextual key
+        :param part: part of the record
+        :return: key to be used in storage's db queries
+        """
+        pass
```
Olegt0rr, line 34: Redundant `pass`. This string may be removed

aiogram/fsm/storage/base.py
```diff
@@ -19,6 +19,68 @@ class StorageKey:
...
+
+
+class DefaultKeyBuilder(KeyBuilder):
+    """
+    Simple key builder with default prefix.
```
Olegt0rr, line 39: Docstring should started without linebreak

```suggestion
    """Simple key builder with default prefix.
```

aiogram/fsm/storage/base.py
```diff
@@ -19,6 +19,68 @@ class StorageKey:
...
+            parts.append(str(key.bot_id))
+        parts.append(str(key.chat_id))
+        if key.thread_id:
+            parts.append(str(key.thread_id))
+        parts.append(str(key.user_id))
+        if self.with_destiny:
+            parts.append(key.destiny)
+        elif key.destiny != DEFAULT_DESTINY:
+            raise ValueError(
+                "Default key builder is not configured to use key destiny other than the default.\n"
```
Olegt0rr, line 76: Please, avoid direct passing string to an error to avoid traceback output duplications. Assign string to variable first.

aiogram/fsm/storage/mongo.py
```diff
@@ -0,0 +1,118 @@
...
+        self._db_name = db_name
+        self._states_collection: AsyncIOMotorCollection = self._client[db_name][
+            states_collection_name
+        ]
+        self._data_collection: AsyncIOMotorCollection = self._client[db_name][data_collection_name]
+        self._key_builder = key_builder
+
+    @classmethod
+    def from_url(
+        cls, url: str, connection_kwargs: Dict[str, Any] = {}, **kwargs: Any
```
Olegt0rr, line 47: Don't create mutable objects in default args

aiogram/fsm/storage/mongo.py
Olegt0rr: 2. https://www.mongodb.com/community/forums/t/how-to-update-a-nested-document-use-operator-in-mongodb/170700

aiogram/fsm/storage/base.py
Olegt0rr: Ah, other code uses the same format. Skipped.

aiogram/fsm/storage/base.py
Rishat-F: `pass` is used within `@abstractmethod` in current codebase in many places. Therefore, it's better to leave it, so that everything is uniform.
If it should be removed from every module in codebase, seems like it's needed separate Pull Request with refactoring.

aiogram/fsm/storage/mongo.py
Rishat-F: 2. Finally find how to do that - https://www.mongodb.com/community/forums/t/upsert-nested-object/204990

Rishat-F: 1. @Olegt0rr, I meant the following difficulty for the case if we store state and data together.
State was created:
`{
    _id: "fsm:bot_id:user_id:chat_id:destiny", 
    state: state
}`
Then data was created:
`{
    _id: "fsm:bot_id:user_id:chat_id:destiny", 
    state: state,
    data: {...}
}`
When state (data) would be removed we need to check if data (state) is presented in document. If data (state) not presented in document we should remove whole document, because no reason to store document with only "_id" key, like this:
`{
    _id: "fsm:bot_id:user_id:chat_id:destiny"
}`
I thought that the implementation of this behavior requires more code and additional logic.
But now i see, that maybe it would be not so complicated as I thought. Okay, I'll give it a try.

aiogram/fsm/storage/mongo.py
```diff
@@ -0,0 +1,120 @@
...
+        :param client: Instance of AsyncIOMotorClient
+        :param key_builder: builder that helps to convert contextual key to string
+        :param db_name: name of the MongoDB database for FSM
+        :param states_collection_name: name of the collection for storing FSM states.
+        :param data_collection_name: name of the collection for storing additional data.
+        """
+        if key_builder is None:
+            key_builder = DefaultKeyBuilder()
+        self._client = client
+        self._db_name = db_name
```
Olegt0rr, line 38: `self._db_name` - seems redundant

aiogram/fsm/storage/mongo.py
```diff
@@ -0,0 +1,120 @@
...
+        self._states_collection: AsyncIOMotorCollection = self._client[db_name][
+            states_collection_name
+        ]
+        self._data_collection: AsyncIOMotorCollection = self._client[db_name][data_collection_name]
```
Olegt0rr, line 42: Get single database instance via `self._client[db_name]` first, then reuse it for getting collection objects.
It will make code easier to read

aiogram/fsm/storage/mongo.py
Rishat-F: It is used in tests teardown

aiogram/fsm/storage/mongo.py
Rishat-F: Done

JrooTJunior (changes requested):

.github/workflows/tests.yml
```diff
@@ -104,7 +108,7 @@ jobs:
       matrix:
         os:
           - ubuntu-latest
-          - macos-latest
+          # - macos-latest
```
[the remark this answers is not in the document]
Rishat-F: What is the difference between `tests` and `pypy-tests` sections in workflows/tests.yml? Seems like `pypy-tests `section does not needed, because tests already runs in `tests` section.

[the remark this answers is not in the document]
JrooTJunior: Separated because additional conditions in the job makes it harder to read and maintain, separated workflows easiest way to manage it.

Rishat-F:
@JrooTJunior, what is the status of PR? Do you wait something from me to do?

JrooTJunior:
> Do you wait something from me to do?

Waiting for free time to review and test it before merge.

Vadim-Khristenko:
Hi all,
@Rishat-F Can you merge Aiogram 3.5.0, which adds full support for Bot API 7.2 with your PR? It would be cool because your PR allows me to use MongoDB in FSM and not have to resort to installing Redis or anything else. But unfortunately, in your PR you only have Bot API 7.1, and Bot API 7.2 and Aiogram 3.5.0 have already been released.

@JrooTJunior If possible, can you tell me the approximate date when the merger will take place, if at all. It would just be nice to see this, for example, in 3.5.1 or just in the near future.

Rishat-F:
@Vadim-Khristenko I'll try it in a couple of weeks.

Rishat-F:
@Vadim-Khristenko aiogram 3.5.0 merged in this PR

Rishat-F:
@JrooTJunior could we complete this PR?

JrooTJunior (approved):

PR #1506, opened 2024-06-10
`InputFile` reuse recommendations update

beliboba:
who the fuck even wrote this

# Description

### DIFF
--- a/docs/api/upload_file.rst
+++ b/docs/api/upload_file.rst
@@ -20,12 +20,11 @@ Here are the three different available builtin types of input file:
 
 .. warning::
 
-    **Be respectful with Telegram**
+    **Be respectful to Telegram**
 
     Instances of `InputFile` are reusable.
-    That's mean you can create instance of InputFile and sent this file multiple times but Telegram
-    does not recommend to do that and when you upload file once just save their `file_id`
-    and use it in next times.
+    That means you can create multiple instances of InputFile for the same file and send them each but Telegram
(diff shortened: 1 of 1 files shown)


beliboba:
im not doing allat mf😂

docs/api/upload_file.rst
```diff
@@ -20,12 +20,11 @@ Here are the three different available builtin types of input file:
...
 
-    **Be respectful with Telegram**
+    **Be respectful to Telegram**
 
     Instances of `InputFile` are reusable.
-    That's mean you can create instance of InputFile and sent this file multiple times but Telegram
-    does not recommend to do that and when you upload file once just save their `file_id`
-    and use it in next times.
+    That means you can create multiple instances of InputFile for the same file and send them each but Telegram
+    does not recommend it. Instead you can send one instance of InputFile for the same picture and send it multiple times using its `file_id`.
```
Olegt0rr, line 27: It's not clear that you should save `file_id` from the first sent instance

Olegt0rr (changes requested):

docs/api/upload_file.rst
akchonya, line 27:
```suggestion
    That means you can create an instance of InputFile and send it multiple times. However, Telegram does not recommend doing this. Instead, once you upload a file, save its `file_id` and reuse that later.
```

akchonya (changes requested):

PR #1510, opened 2024-06-13
Fail redis and mongo tests if incorrect URI provided + some storages tests refactoring

Rishat-F:
# Description

If incorrect URIs provided to "**--redis**" and/or "**--mongo**" options tests should fail with **errors** instead of **skipping**.
Otherwise the next scenario is possible:

1. developer breaks RedisStorage and/or MongoStorage code
2. tests are run with incorrect redis and/or mongo URIsprovided by "**--redis**" and "**--mongo**" options (for example, wrong port specified)
3. tests pass because skipping doesn't fail tests run
4. developer or reviewer doesn't notice that redis and/or mongo tests were skipped
5. broken code gets in codebase

# Checklist:


Olegt0rr (changes requested):
LGTM, but make 100% coverage, please

aiogram/fsm/storage/mongo.py
```diff
@@ -39,7 +39,10 @@ def __init__(
 
     @classmethod
     def from_url(
-        cls, url: str, connection_kwargs: Optional[Dict[str, Any]] = None, **kwargs: Any
+        cls,
+        url: str,
+        connection_kwargs: Optional[Dict[str, Any]] = {"serverSelectionTimeoutMS": 2000},
```
Olegt0rr, line 44: Why you pass this values as default?

Rishat-F: Yes, I agree. That was bad idea to do like this.

Rishat-F:
> LGTM, but make 100% coverage, please

Done

examples/multibot.py
```diff
@@ -4,6 +4,7 @@
 from typing import Any, Dict, Union
 
 from aiohttp import web
+from finite_state_machine import form_router
```
Olegt0rr, line 7: Looks like it's conflicts with command `make reformat` from contribution guide

tests/conftest.py
```diff
@@ -5,19 +5,27 @@
...
+from aiogram.fsm.storage.redis import RedisStorage
 from tests.mocked_bot import MockedBot
 
 DATA_DIR = Path(__file__).parent / "data"
 
+CHAT_ID = -42
+USER_ID = 42
+
+skip_message_pattern = 'Need "--{db}" option with {db} URI to run'
+invalid_uri_pattern = "Invalid {db} URI {uri!r}: {err}"
```
Olegt0rr, line 27: Looks like CONSTANT vars

Olegt0rr (changes requested):

tests/conftest.py
Rishat-F: Done

examples/multibot.py
Rishat-F: Nothing had changed after running `make reformat`

Olegt0rr: Please update and merge `aiogram:dev-3.x` to this branch.
We're just updated dependencies of the format checkers and linters.

Rishat-F: Done

Olegt0rr (approved):

JrooTJunior (approved):
Thanks for your contribution.

I will update the changelog message and category, this change does not affect the functionality of the framework, but improves the development process of the framework itself, so these changes should be marked as "misc" and not as "bugfixes".

PR #1522, opened 2024-06-22
Added functionality for negative numbers in keyboard adjust.

N0rmalUser:
# Description

Added functionality for negative numbers in keyboard adjust.

## Type of change
Если функция получает отрицательное число в `adjust`, она применяет их к нижним кнопкам так же, как и положительные числа влияют на верхние кнопки. Отрицательные числа имеют больший приоритет в форматировании.

# How Has This Been Tested?

```
def test_markup():
    builder = InlineKeyboardBuilder()
    for i in range(12):
        builder.button(text=f"test{i}", callback_data=f"test{i}")
    builder.adjust(3, 1, -2)
    return builder.as_markup()

@router.message(Command('test'))
async def test_handler(msg: Message) -> None:
    await msg.answer("This is TEST!!!", reply_markup=test_markup())
```

```
def test_markup():
    builder = InlineKeyboardBuilder()
    for i in range(1):
        builder.button(text=f"test{i}", callback_data=f"test{i}")
    builder.adjust(-2)
    return builder.as_markup()
```

```
def test_markup():
    builder = ReplyKeyboardBuilder()
    builder.row(*(KeyboardButton(text=f"test-{index}") for index in range(6)))
    builder.button(text="Add", callback_data="add")
    builder.button(text="Cancel", callback_data="cancel")
    builder.adjust(3, -1, -1)
    return builder.as_markup()
```

**Test Configuration**:
* Python version: 3.11.9

# Checklist:


JrooTJunior (changes requested):
This feature is very cool, I thought about such functionality for a long time (about 1.5 year ago) and it's in my deep backlog with low priority, thanks for trying to implement it.

We currently have an alternative way to do this using a combination of constructors (added in v3.0.0rc1 #1236), but it's not the same as what you're doing here.

So, regarding your solution, I think the design you chose is a little strange because it's not intuitive. I have another idea of ​​what it might look like:

```python
builder = InlineKeyboardBuilder()
# ... add buttons

builder.adjust(2, 3, ..., 5)
```

In this example, `...` (circular object) is used as a separator between parts, it should be specified only once per configuration and means that the left part of the values ​​should be used for the first part of the markup, and then the right part should be used for the last part of the markup.
Technically, we can find the position of the circular object in the size list and split the positions for the first and last pieces by that configuration.

I think this is more intuitive than negative values.

Let's discuss it.

tests/test_utils/test_keyboard.py
```diff
@@ -183,11 +183,17 @@ def test_row(
...
+            [3, False, [-2], [1, 2]],
             [10, False, [3, 2, 1], [3, 2, 1, 1, 1, 1, 1]],
+            [10, False, [3, 2, -1], [3, 2, 2, 2, 1]],
+            [10, False, [3, -2, 1], [3, 1, 1, 1, 1, 1, 2]],
             [12, False, [], [10, 2]],
             [12, True, [3, 2, 1], [3, 2, 1, 3, 2, 1]],
+            [12, True, [3, -1, -4], [3, 3, 1, 1, 4]],
         ],
     )
     def test_adjust(self, count, repeat, sizes, shape):
```
JrooTJunior, line 199: What if we have fewer buttons than the total of all sizes?

PR #1523, opened 2024-06-23
Fixed webhook Bot initialization to use default properties in updated…

GaniyevUz:
# Description

This PR updates the Bot initialization to comply with the changes introduced in aiogram version 3.7.0. The `parse_mode`, `disable_web_page_preview`, and `protect_content` arguments have been removed from the Bot initializer. Instead, the `default=DefaultBotProperties(parse_mode=<ParseMode.HTML: 'HTML'>)` argument is now used to set default properties.

## Type of change

# How Has This Been Tested?

- [x] Manual 
- [x] Verified that the bot initializes correctly with the new `DefaultBotProperties` argument.
- [x] Tested webhook handling to ensure it functions as expected with the updated aiogram version.

**Test Configuration**:
* Operating System: Ubuntu 22.04
* Python version: 3.10

# Checklist:

### DIFF
--- a/aiogram/webhook/aiohttp_server.py
+++ b/aiogram/webhook/aiohttp_server.py
@@ -14,6 +14,7 @@ from aiogram.methods import TelegramMethod
 from aiogram.methods.base import TelegramType
 from aiogram.types import InputFile
 from aiogram.webhook.security import IPFilter
+from aiogram.client.default import DefaultBotProperties
 
 
 def setup_application(app: Application, dispatcher: Dispatcher, /, **kwargs: Any) -> None:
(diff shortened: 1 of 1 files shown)


JrooTJunior:
This solution is completely wrong, you don't need to replace the type `bot_settings` from `dict` with a specific class `DefaultBotProperties` because you:
1. You made it impossible to specify the session (`bot_settings={"session": my_custom_session_with_proxy}`)
2. You should just pass `bot_settings={"default": DefaultBotProperties(parse_mode=ParseMode.HTML)}` instead of `bot_settings={"parse_mode": ParseMode.HTML}`.

So there is no broken feature that you are trying to fix.

PR #1546, opened 2024-07-25
Fix web_app example

bogdan-nikitin:
# Description

Fixed URL path in the "Open" button at the "demo/sendMessage" endpoint in the web_app example. 
The button should open the web application, but at the moment it sends a GET request to the current endpoint (which only accepts POST requests).

## Type of change

### DIFF
--- a/CHANGES/1538.misc.rst
+++ b/CHANGES/1538.misc.rst
@@ -1,3 +1,0 @@
-Added method :func:`aiogram.types.message.Message.as_reply_parameters`.
-Replaced usage of the argument :code:`reply_to_message_id` with :code:`reply_parameters`
-in all Message reply methods.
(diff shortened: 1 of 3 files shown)


CHANGES/1538.misc.rst
```diff
@@ -1,3 +0,0 @@
-Added method :func:`aiogram.types.message.Message.as_reply_parameters`.
```
JrooTJunior, line 1: You shouldn't delete this file.

JrooTJunior (changes requested):

CHANGES/1538.misc.rst
bogdan-nikitin: Restored

JrooTJunior (approved):

PR #1564, opened 2024-08-17
Fix readable BufferedInputFile and add support BufferedIOBase

Teri-anric:
# Description

- Fixed not safe read file of method BufferedInputFile.from_file.
- Added support initialization BufferedInputFile from BytesIO or other bytes buffer's.

## Type of change

# How Has This Been Tested?

- [x] Test by unit tests

**Test Configuration**:
* Operating System: Win-11 x64
* Python version: 3.8

# Checklist:

### DIFF
--- a/CHANGES/1564.misc.rst
+++ b/CHANGES/1564.misc.rst
@@ -1,0 +1,5 @@
+Fixed unsafe file reading in BufferedInputFile.from_file and 
+added support for initializing BufferedInputFile from BytesIO and 
+other bytes buffers. 
+This change improves safety and extends flexibility in handling 
+different byte sources when working with buffered input files.
--- a/aiogram/types/input_file.py
+++ b/aiogram/types/input_file.py
@@ -38,7 +38,12 @@ class InputFile(ABC):
 
 
 class BufferedInputFile(InputFile):
-    def __init__(self, file: bytes, filename: str, chunk_size: int = DEFAULT_CHUNK_SIZE):
+    def __init__(
+        self,
+        buffer_or_bytes: bytes | io.BufferedIOBase,
+        filename: str,
+        chunk_size: int = DEFAULT_CHUNK_SIZE,
+    ):
         """
         Represents object for uploading files from filesystem
 
@@ -48,7 +53,10 @@ class BufferedInputFile(InputFile):
         """
         super().__init__(filename=filename, chunk_size=chunk_size)
 
-        self.data = file
+        if isinstance(buffer_or_bytes, bytes):
+            buffer_or_bytes = io.BytesIO(buffer_or_bytes)
+
+        self.buffer = buffer_or_bytes
 
     @classmethod
     def from_file(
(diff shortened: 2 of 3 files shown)


CHANGES/1564.misc.rst
```diff
@@ -0,0 +1,5 @@
+Fixed unsafe file reading in BufferedInputFile.from_file and
```
JrooTJunior, line 1: What does "unsafe" mean?

--- 

UPD: This comment was the first in the overview of the changes, but after reading all the changes it became clear to me that with these changes you on the contrary make it completely unsafe to use.

aiogram/types/input_file.py
```diff
@@ -38,7 +38,12 @@ async def read(self, bot: "Bot") -> AsyncGenerator[bytes, None]:  # pragma: no c
 
 
 class BufferedInputFile(InputFile):
-    def __init__(self, file: bytes, filename: str, chunk_size: int = DEFAULT_CHUNK_SIZE):
+    def __init__(
+        self,
+        buffer_or_bytes: bytes | io.BufferedIOBase,
```
JrooTJunior, line 43: The "file" argument should not be renamed because someone could use named arguments and this change breaks compatibility.

---

The `A | B` type hint is not compatible with Python 3.8.

---

`SrtingIO` is also a subclass of `BufferedIOBase`, but it is incompatible with the `InputFile.read` method, which should only generate chunks of bytes.

aiogram/types/input_file.py
```diff
@@ -68,13 +76,15 @@ def from_file(
...
         if filename is None:
             filename = os.path.basename(path)
-        with open(path, "rb") as f:
-            data = f.read()
-        return cls(data, filename=filename, chunk_size=chunk_size)
+        buffer = io.BytesIO()
+        with open(path, "rb") as fp:
+            while chunk := fp.read(chunk_size):
+                buffer.write(chunk)
+        return cls(buffer, filename=filename, chunk_size=chunk_size)
```
JrooTJunior, line 83: This modification didn't change anything because `from_file` is still blocking.

Instead of this change, I see the sense of adding another method called `load_from_file`, which should be asynchronous and use `aiofiles` to read the contents into the bytes variable.

Note that both methods should be here because you can initialize `InputFile` globally to reuse a static file without performing disk I/O operations.

aiogram/types/input_file.py
```diff
@@ -68,13 +76,15 @@ def from_file(
...
 
     async def read(self, bot: "Bot") -> AsyncGenerator[bytes, None]:
-        buffer = io.BytesIO(self.data)
-        while chunk := buffer.read(self.chunk_size):
+        self.buffer.seek(0)
+        while chunk := self.buffer.read(self.chunk_size):
```
JrooTJunior, line 87: This change makes BufferedInputFile unsafe for concurrency because when you try to send the same InputFile from different coroutines, you're trying to read from the same BytesIO object.

Example:
1. Coroutine 1 read chunk 1
2. Coroutine 2 read chunk 2
3. Coroutine 1 read chunk 3

After that, both coroutines send the corrupted file.

JrooTJunior (changes requested):

PR #1577, opened 2024-09-10
Bump aiofiles version upper bound to <24.2

monosans:
I think the upper bound can be removed altogether, since aiogram only uses aiofiles.open, which is unlikely to ever be changed. Let me know what you think is best.


Olegt0rr:
LGTM

PR #1596, opened 2024-10-22
Don't use KeyboardBuilder directly

### LINKED ISSUE #1595: KeyboardBuilder: if self.max_width = 0, endless while loop
### Checklist

- [X] I am sure the error is coming from aiogram code
- [X] I have searched in the issue tracker for similar bug reports, including closed ones

### Operating system

Linux

### Python version

3.11, 3.12

### aiogram version

3.13.1

### Expected behavior

Keyboard builder is created.

### Current behavior

```
        # Separate buttons to exclusive rows with max possible row width
        while buttons:
            row, buttons = buttons[: self.max_width], buttons[self.max_width :]
            markup.append(list(row))
```

This code results in an endless loop if self.max_width = 0, since the number of buttons never decreases.

### Steps to reproduce

```
        builder = KeyboardBuilder(button_type=KeyboardButton)
        for device_id in devices:
            button = KeyboardButton(text=device_id)
            builder.add(button)
```

When max_width is explicitly set, code runs without an issue.
```
        builder = KeyboardBuilder(button_type=KeyboardButton)
        builder.max_width = 5
        for device_id in devices:
            button = KeyboardButton(text=device_id)
            builder.add(button)
```

skinkie:
As discussed in issue #1595 KeyboardBuilder should not be used directly. Instead InlineKeyboardBuilder or ReplyKeyboardBuilder should be used. This fix makes sure that the documentation rendered does not include a "wrong example".

# Description

Make the example a developer best practise.
 
Fixes #1595 

## Type of change

- [X] Documentation (typos, code examples or any documentation update)
- [X] This change requires a documentation update

# How Has This Been Tested?

# Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have made corresponding changes to the documentation

### DIFF
--- a/aiogram/utils/keyboard.py
+++ b/aiogram/utils/keyboard.py
@@ -152,7 +152,7 @@ class KeyboardBuilder(Generic[ButtonType], ABC):
 
         .. code-block:: python
 
-            >>> builder = KeyboardBuilder(button_type=InlineKeyboardButton)
+            >>> builder = InlineKeyboardBuilder(button_type=InlineKeyboardButton)
             >>> ... # Add buttons to builder
             >>> markup = InlineKeyboardMarkup(inline_keyboard=builder.export())
 


aiogram/utils/keyboard.py
```diff
@@ -152,7 +152,7 @@ def export(self) -> List[List[ButtonType]]:
 
         .. code-block:: python
 
-            >>> builder = KeyboardBuilder(button_type=InlineKeyboardButton)
+            >>> builder = InlineKeyboardBuilder(button_type=InlineKeyboardButton)
```
JrooTJunior, line 155: Since this line of documentation is shown in all child classes, this change does not make this example any less confusing for all other cases, so I disagree with this change.

- https://docs.aiogram.dev/en/stable/utils/keyboard.html#aiogram.utils.keyboard.InlineKeyboardBuilder.export
- https://docs.aiogram.dev/en/stable/utils/keyboard.html#aiogram.utils.keyboard.ReplyKeyboardBuilder.export

Let's find another way.

JrooTJunior (changes requested):

aiogram/utils/keyboard.py
Olegt0rr: Now docs works the same way `InlineKeyboardMarkup` displayed in the `ReplyKeyboard` example

Olegt0rr: Or should we make `KeyboardBuilder` abstract?

PR #1609, opened 2024-11-27
Fixed the order of handlers in the Scene

RootShinobi:
Fixed the order of handlers in the Scene


JrooTJunior (changes requested):
This change does not fix the mentioned bug because `dir` returns an ordered list of members just like `inspect.getmembers`, so we need to find another way to solve this problem.

PR #1616, opened 2024-12-12
Added __call__ method to TelegramMethod class.

qvvonk:
# Description

Added `__call__` method to the `aiogram.methods.base.TelegramMethod` class.

The exception text in the `aiogram.methods.base.TelegramMethod.__await__` method reads:

> This method is not mounted to any bot instance, please call it explicilty with bot instance `await bot(method)` or mount method to a bot instance `method.as_(bot)` and then call it `await method()`

But if you try to use `await TelegramMethod()`, you will get an exception: 
> TypeError: <TelegramMethod> object is not callable

In turn, using `await TelegramMethod` without `()` is a bit confusing and doesn't match the hint in the exception text.

## Type of change

# How Has This Been Tested?

# Checklist:

- [ ]  I have made corresponding changes to the documentation
- [ ]  New and existing unit tests pass locally with my changes


JrooTJunior (changes requested):
I do not agree with these changes. We shouldn't add more ways to do the same thing.

Looks like a typo in the error message, so just remove the brackets from the error message instead of adding this method.

gygabrain:
Done.
But, don't you find the use of `await` in conjunction with name without calling (`()`) unintuitive?
From the very beginning of learning asynchronous programming (at least in _Python_), people are told that the object call itself returns a coroutine, rather than starting to execute any logic.
On the other hand, replacing the `__await__` method with` __call__` just like that right away is not a good option either, since there are probably many people already using the existing option.
Wouldn't it be appropriate to leave the `__await__` method deprecated and add the `__call__` method?
Or is there some special reason for using `__await__` specifically?

JrooTJunior (approved):

PR #1628, opened 2025-01-09
fix: change the `Downloadable` protocol to be non-writable to shut up type checking that checks code that uses the `bot.download(...)` method

monosans:
The following code gives error with type-checkers, because `TelegramObject` is frozen and `Downloadable` Protocol expects `file_id` to be writable:

```python
from aiogram import Bot
from aiogram.types import Document


async def main() -> None:
    async with Bot("") as bot:
        await bot.download(Document(file_id="", file_unique_id=""))
```

```python
error: Argument 1 to "download" of "Bot" has incompatible type "Document"; expected "str | Downloadable"  [arg-type]
note: Protocol member Downloadable.file_id expected settable variable, got read-only attribute
```

## Type of change


Olegt0rr (approved):

CHANGES/1628.bugfix.rst
```diff
@@ -0,0 +1 @@
+Don't require :code:`Downloadable.file_id` to be writable
```
JrooTJunior, line 1: I think this variant should be more understandable for other developers

```suggestion
Change the :code:`Downloadable` protocol to be non-writable to shut up type checking that checks code that uses the :code:`bot.download(...)` method
```

JrooTJunior (changes requested):

Issue #1641, opened 2025-02-16
Bug: Incorrect Ordering of Registered Handlers in `Scene` Object


JrooTJunior:
When registering handlers in the `Scene` object, the order of handlers can become incorrect due to the behavior of `inspect.getmembers`. This Python standard library function [returns members sorted alphabetically](https://docs.python.org/3/library/inspect.html#inspect.getmembers), which disrupts the intended sequence of the handlers.

This creates a problem when a handler with broader filters comes before more specific ones, leading to the broader handler being executed before the stricter ones. This behavior is not intuitive and can lead to undesired runtime outcomes.

The following code snippet demonstrates an affected area:
https://github.com/aiogram/aiogram/blob/21c74c73d07b0d170779cac4dbac4646e686021d/aiogram/fsm/scene.py#L326-L340

In this logic:
- The `inspect.getmembers(cls)` call retrieves all members of the class sorted alphabetically by name.
- Any handlers with broader filter rules (that happen to appear alphabetically earlier by name) are registered before others with stricter rules.

## Expected Behavior
Handlers should be registered in their **defined order** within the `Scene` class rather than an alphabetical order. This would ensure that specific handlers are executed prior to broader ones when their execution depends on filtering precedence.

## Suggested Fix
Instead of `inspect.getmembers`, consider alternatives that preserve the order of definition in the class:
- Use Python's `__dict__` or `__dir__` for inspecting class attributes, which may retain the definition order.
- Leverage [PEP 520](https://peps.python.org/pep-0520/) introduced in Python 3.6, where class attribute definition order is explicitly preserved.

## Steps to Reproduce
1. Define handlers in a `Scene` subclass where a broader handler (e.g., no or few filters) comes alphabetically before a stricter one (e.g., with more filters).
2. Observe that the broader handler is executed first, even though it may be intended to run later.

## Impact
This issue can lead to:
- Unexpected handler execution behavior.
- Difficult-to-debug runtime issues, particularly in complex scenes with multiple handlers.

PR #1642, opened 2025-02-16
Fix handler registration order in `Scene`

JrooTJunior:
Previously, `Scene` handlers were registered based on the sorted output of `inspect.getmembers`, causing incorrect execution order. Now, handlers are registered in the order they are defined in the class, ensuring reliable behavior and proper sequence when handling filters with varying specificity. Added test cases to validate the correct handler ordering.

Fixes: #1641


andrew000:
Works correct in my project

PR #1668, opened 2025-04-05
CallbackData with default prefix

Coderik:
# Description

This PR modifies `CallbackData` and makes the prefix optional. By default the class name is used as prefix.

This makes it less tedious to define a large number of custom `CallbackData` classes. For instance:

```python
class SettingsActions:
    class GoBack(CallbackData):
        pass

    class ShowFirstMenu(CallbackData):
        pass

    class ShowSecondMenu(CallbackData):
        pass

    class SetSomeFlag(CallbackData):
        flag: bool
```

**NOTE:** This PR does not touch the documentation, because I was not sure how welcome this change would be. It made my own code easier to maintain. If you think it's useful, I'll update the documentation.

## Type of change

# How Has This Been Tested?

This PR removes one test that checks that an exception is raised when the prefix is not provided. And adds one test that checks the new behavior.

- [x] Test that the prefix is set to class name, when not given explicitly
- [x] Test that the prefix can be set explicitly

**Test Configuration**:
* Operating System: Linux
* Python version: 3.10.12

# Checklist:

### DIFF
--- a/aiogram/filters/callback_data.py
+++ b/aiogram/filters/callback_data.py
@@ -47,24 +47,26 @@ class CallbackData(BaseModel):
 
     This class should be used as super-class of user-defined callbacks.
 
-    The class-keyword :code:`prefix` is required to define prefix
-    and also the argument :code:`sep` can be passed to define separator (default is :code:`:`).
+    An optional class-keyword :code:`prefix` can be passed to define prefix.
+    If no prefix is provided, the class name will be used as the prefix.
+
+    An optional argument :code:`sep` can be passed to define the separator
+    (default is :code:`:`).
     """
 
     if TYPE_CHECKING:
         __separator__: ClassVar[str]
         """Data separator (default is :code:`:`)"""
         __prefix__: ClassVar[str]
-        """Callback prefix"""
+        """Callback prefix (default is class name)"""
 
     def __init_subclass__(cls, **kwargs: Any) -> None:
-        if "prefix" not in kwargs:
-            raise ValueError(
-                f"prefix required, usage example: "
-                f"`class {cls.__name__}(CallbackData, prefix='my_callback'): ...`"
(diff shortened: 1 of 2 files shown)


aiogram/filters/callback_data.py
```diff
@@ -47,24 +47,26 @@ class CallbackData(BaseModel):
...
-                f"prefix required, usage example: "
-                f"`class {cls.__name__}(CallbackData, prefix='my_callback'): ...`"
-            )
+        # If no prefix is provided explicitly, default to the class name
+        prefix = kwargs.pop("prefix", None)
+        if prefix is None:
+            prefix = cls.__name__
         cls.__separator__ = kwargs.pop("sep", ":")
-        cls.__prefix__ = kwargs.pop("prefix")
+        cls.__prefix__ = prefix
```
Vadim-Khristenko, line 69: Why not simply do this?
`cls.__prefix__ = kwargs.pop("prefix", cls.__name__)`

JrooTJunior:
Please provide more details and explain why these changes are needed.

Please note that callback data has a strict size limit (64 bytes) on the Telegram Bot API side, so you should not increase the data size. In most cases, 1-5 characters are enough for the prefix part.

But on the other hand, class names should be as clear and self-descriptive as possible, so in most cases this means that there should be more than 5 characters to explain what this class does (Exactly what you showed in your example).

Therefore, these statements are mutually exclusive.

Coderik:
> Please provide more details and explain why these changes are needed.

My use case is a fairly large, a couple of levels deep settings menu, built on inline keyboards. I use CallbackData classes as actions to move between levels and adjust settings. In this case three are many CallbackData classes either without, or with a single data field. So, there is no issues with the 64 bytes limit. But I'm forced to define the prefix anyways.

PR #1680, opened 2025-04-27
Fix: Add alias for `amount` field in `StarAmount` to match API response (star_count)

flvbs7:
# Description

Fixed the issue with the mapping of the `StarAmount` type  to match the Telegram API response. The issue occurred when trying to retrieve the amount of stars, as the field `amount` was missing in the response, while Telegram returns `star_count`. 

## Type of change

# How Has This Been Tested?

Called the `get_business_account_star_balance` method and verified that the returned `StarAmount` object now correctly maps the `star_count` field to the `amount` attribute.

- [x] Test with actual Telegram bot API to ensure the fix works as expected

**Test Configuration**:
* Python version: 3.13.1

# Checklist:

### DIFF
--- a/CHANGES/1680.bugfix.rst
+++ b/CHANGES/1680.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed an issue in the :code:`StarAmount` type where the field :code:`amount` was not correctly deserialized when retrieving the business account star balance.
+The response from the Telegram API contains the field `star_count`, which now correctly maps to the `amount` field in the `StarAmount` class.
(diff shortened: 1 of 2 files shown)


JrooTJunior:
There seems to be an error in the documentation or the Bot API.

I have notified the Bot API developers of this inconsistency, so until the API or documentation is fixed we should not change anything in the framework.

telegram-cloud-photo-size-2-5219735710188302119-y

Also in case when we have using code generation for Bot API methods and types, we shouldn't change anything by hands at this place because it will be regenerated soon.

PR #1705, opened 2025-07-09
Migrate motor to pymongo

### LINKED ISSUE #1695: Migrate from Motor to PyMongo Async for MongoStorage
### aiogram version

3.x

### Problem

Motor is deprecated now, we should use pymongo async.

### Possible solution

https://www.mongodb.com/docs/languages/python/pymongo-driver/current/reference/migration/

### Alternatives

_No response_

### Code example

```python3

```

### Additional information

_No response_

kievzenit:
# Description

Migrated `MongoStorage` from relying on deprecated `motor` package to using new async `PyMongo`. Also `__init__.py` file was changed in `aiogram/storage/fsm` to improve developer experience.

Fixes #1695 

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

It was tested utilizing method that is provided in contributing guide: `pytest --mongo mongodb://<user>:<password>@<host>:<port> tests`

**Test Configuration**:
* Operating System: Linux fedora 6.15.4-100.fc41.x86_64 x86_64 GNU/Linux
* Python version: 3.13.5
* MongoDB version: 6.0.22-ubi8

# Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,7 +64,7 @@ redis = [
 ]
 mongo = [
     "motor>=3.3.2,<3.7.0",
-    "pymongo>4.5,<4.10",
+    "pymongo>4.5,<4.11",
 ]
 proxy = [
     "aiohttp-socks~=0.8.3",


aiogram/fsm/storage/__init__.py
```diff
@@ -0,0 +1,11 @@
+from .base import BaseStorage
+from .memory import MemoryStorage
+from .mongo import MongoStorage
+from .redis import RedisStorage
+
+__all__ = (
```
JrooTJunior, line 6: This change will break the import of this package because the Redis and Mongo drivers are in optional dependencies and are not installed in most cases, so importing any repository from this package may result in an `ImportError` in all cases where the optional dependencies are not installed.

JrooTJunior (changes requested):
This is a very small but critical change for developers using this type of storage, as it could break code until the developer rewrites other code that uses Mongo. Also, using multiple drivers for the same purposes can be too expensive.

So, the best way to update this storage is to mark the current implementation as deprecated and add a new storage implementation with a name like `PyMongoStorage`, and provide migration guidance in the changelog and documentation articles (storages section).

JrooTJunior (changes requested):

aiogram/fsm/storage/pymongo.py
```diff
@@ -0,0 +1,136 @@
...
+        :return: an instance of :class:`PyMongoStorage`
+        """
+        if connection_kwargs is None:
+            connection_kwargs = {}
+        client = AsyncMongoClient(url, **connection_kwargs)
+        return cls(client=client, **kwargs)
+
+    async def close(self) -> None:
+        """Cleanup client resources and disconnect from MongoDB."""
+        self._client.close()
```
Vadim-Khristenko, line 60: Hi! Look, you have a problem here - you forgot to add `await`. This method in Async PyMongo should be awaited.

```python
async def close(self) -> None:
    """Cleanup client resources and disconnect from MongoDB."""
    await self._client.close()
```

Vadim-Khristenko, line 60: Hi! Look, you have a problem here - you forgot to add `await`. This method in Async PyMongo should be awaited.

```python
async def close(self) -> None:
    """Cleanup client resources and disconnect from MongoDB."""
    await self._client.close()
```

kievzenit: Yeah, my bad. I fixed it and added a test that checks if client could be closed correctly.

Vadim-Khristenko (approved):

kievzenit:
Copilot has pointed out an actual problem in the code and as far as I just copied old implementation of `MongoStorage` to new one `PyMongoStorage`, because there were no change in calling interface, this issue is still present in `MongoStorage` class. How should I address it? Create new PR and fix it there or fix it in this PR? @JrooTJunior @Vadim-Khristenko

aiogram/fsm/storage/pymongo.py
```diff
@@ -0,0 +1,136 @@
...
+        update_with = {f"data.{key}": value for key, value in data.items()}
+        update_result = await self._collection.find_one_and_update(
+            filter={"_id": document_id},
+            update={"$set": update_with},
+            upsert=True,
+            return_document=True,
+            projection={"_id": 0},
+        )
+        if not update_result:
+            await self._collection.delete_one({"_id": document_id})
```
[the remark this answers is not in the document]
kievzenit: Actually it will be true. Because of `projection={"_id": 0}`. When there is no record in database and use tries to call `update_data(key, data={})`, `update_result` will be `{}`. So this condition will be true.

kievzenit:
Some of the tests fails are transient, they should go on tests rerun.

Vadim-Khristenko (approved):
Everything looks good — great work! 

Just one suggestion: I recommend pinning the upper version of `pymongo` to `<4.11` for now.  

**Why?**  
- `pymongo==4.10.1` still supports PyPy 3.9.  
- `pymongo>=4.11` drops support for PyPy 3.9 entirely.  
 
Since we currently run tests on `pypy3.9`, upgrading beyond 4.10.1 would break CI.  
 
**Suggestion**:  
Let’s create an issue to track this and automatically upgrade the upper bound when CPython 3.9 reaches EOL (October 2025).  
This way, we’ll be ready to safely bump `pymongo` without breaking builds. 

Ready for final review from @JrooTJunior before merge.

JrooTJunior (approved):

PR #1722, opened 2025-08-22
CallableObject Improvements

### LINKED ISSUE #1721: lambdas and callable objects are not recognized as awaitable handlers
### Checklist

### Operating system

Ubuntu 24.04

### Python version

3.12.3

### aiogram version

3.22.0

### Expected behavior

The dispatcher is supposed to work with sync and async event handlers.
It strives to detect whether supplied handler is async using introspection.
However it does not recognize lambdas and callable object as async (even if they are)

### Current behavior

On first call to unrecognized async handler python emits log message stating that coroutine was never awaited
On subsequent calls nothing is logged. Unrecognized async handlers are not executed in either case

### Steps to reproduce

1. set `BOT_TOKEN` environment variable and run the example
2. send messages with text "1", "2", "3" to the bot driven by the script

### Code example

```python3
import os
import asyncio
import aiogram

from collections.abc import Awaitable
from collections.abc import Callable


bot = aiogram.Bot(token=os.environ["BOT_TOKEN"])
dp = aiogram.Dispatcher()


async def handler(message: aiogram.types.Message) -> None:
    print(message.text)
    if message.text is not None:
        await message.reply(message.text)


class Handler:
    def __init__(self, handler: Callable[[aiogram.types.Message], Awaitable[None]]) -> None:
        self.handler = handler

    async def __call__(self, message: aiogram.types.Message) -> None:
        return await self.handler(message)


dp.message.register(handler, lambda m: m.text == '1')                           # regular WORKS
dp.message.register(lambda message: handler(message), lambda m: m.text == '2')  # lambdas DOESN'T WORK
dp.message.register(Handler(handler), lambda m: m.text == '3')                  # objects DOESN'T WORK


if __name__ == '__main__':
    asyncio.run(dp.start_polling(bot))
```

### Logs

```sh
$ python example.py
1
/usr/lib/python3.12/asyncio/base_events.py:1988: RuntimeWarning: coroutine 'handler' was never awaited
  handle = None  # Needed to break cycles when an exception occurs.
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
/usr/lib/python3.12/asyncio/base_events.py:1988: RuntimeWarning: coroutine 'Handler.__call__' was never awaited
  handle = None  # Needed to break cycles when an exception occurs.
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

### Additional information

_No response_

Enalite:
# Description

Fixes an issue in `CallableObject` and improves its behavior:  
- Callable classes with an `async __call__` method are now correctly detected and awaited.  
- Synchronous callbacks that return coroutines are now properly awaited.  
- Ensures sync functions still run safely in `asyncio.to_thread`.

Fixes #1721

## Type of change

# How Has This Been Tested?

Tested manually with sync and async callables, including callable classes. Verified that:  
- Coroutine-returning sync functions are awaited correctly  
- Sync functions run in `asyncio.to_thread` when necessary  
- No regression in existing async callables  

- [x] Test A: sync function returning a coroutine  
- [x] Test B: callable class with sync `__call__`  

**Test Configuration**:
* Operating System: Any
* Python version: 3.9+

# Checklist:

### DIFF
--- a/CHANGES/1721.bugfix.rst
+++ b/CHANGES/1721.bugfix.rst
@@ -1,0 +1,9 @@
+Bug Fixes
+=========
+
+CallableObject Improvements
+---------------------------
+
+- Correctly detects and awaits coroutine-returning sync callables
+- Supports callable classes with a sync __call__ method
+- Ensures sync functions still run safely in asyncio.to_thread
--- a/aiogram/dispatcher/event/handler.py
+++ b/aiogram/dispatcher/event/handler.py
@@ -26,7 +26,11 @@ class CallableObject:
 
     def __post_init__(self) -> None:
         callback = inspect.unwrap(self.callback)
-        self.awaitable = inspect.isawaitable(callback) or inspect.iscoroutinefunction(callback)
+        self.awaitable = (
+            inspect.isawaitable(callback)
+            or inspect.iscoroutinefunction(callback)
+            or inspect.iscoroutinefunction(getattr(callback, "__call__", None))
+        )
         spec = inspect.getfullargspec(callback)
         self.params = {*spec.args, *spec.kwonlyargs}
         self.varkw = spec.varkw is not None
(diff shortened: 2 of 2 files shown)


Enalite:
You're right, I hadn’t considered that 😅. As you mentioned in #1721, these types of handlers are indeed a bit unusual. I’ll try to find a better way to support them as well. If you think it’s unnecessary, I can go ahead and close this PR.

PR #1724, opened 2025-09-25
Added `dispatcher` key to event data in `dispatcher.feed_update`, removed redundant unpacking of `workflow_data` in `dispatcher.start_polling`.

qvvonk:
# Description

Added `dispatcher key` to event data in `dispatcher.feed_update`, removed redundant unpacking of `workflow_data` in `dispatcher.start_polling`.

I needed to create synthetic `Update` objects and feed them into `Dispatcher.feed_update` (which, according to its docstring, is **the main entry point** for updates). However, I ran into an issue: with such synthetic events, the `dispatcher` key does not propagate to middlewares, filters, or handlers.

After analyzing the flow of the data dictionary, I found the following:

The dictionary is pre-generated in `Dispatcher.start_polling`. It contains `bots`, `dispatcher`, the unpacked `workflow_data`, and unpacked `kwargs`.
This dictionary is then passed as-is to `Dispatcher.feed_update` via:

```
-> Dispatcher._polling(..., **data_dict)
-> Dispatcher._process_update(..., **data_dict)
-> Dispatcher.feed_update(..., **data_dict)
```

Inside `Dispatcher.feed_update`, the data dictionary is reconstructed:
```
{
    **self.workflow_data,
    **kwargs,
    "bot": bot,
}
```
Here, `workflow_data` is unpacked again, but the `dispatcher` key is missing.
As a result, if we try to feed a synthetic event using `Dispatcher.feed_update`, the data dictionary will not contain the `dispatcher` key, making it inaccessible from handlers, filters, or middlewares (unless with pass it into `**kwargs` of `Dispatcher.feed_update`). Moreover, unpacking `workflow_data` twice is unnecessary, since it’s not used anywhere before `Dispatcher.feed_update` —it’s just passed along.

From what I see, there are two possible approaches:

1. Unpack `workflow_data` in `Dispatcher.start_polling` – in this case, any modifications to `workflow_data` during runtime won’t be reflected in middlewares, filters, or handlers (new keys won’t propagate, removed keys will remain). To propagate changes, `Dispatcher.start_polling` would have to be re-executed.

2. Unpack `workflow_data` right before each `Update` is fed into the router system – this allows any runtime modifications to `workflow_data` to be correctly reflected in the keys passed to middlewares, filters, and handlers.

I chose the second approach.

## Type of change

# How Has This Been Tested?

- [x] pytest tests

**Test Configuration**:
* Operating System: Windows 11
* Python version: 3.13

# Checklist:


aiogram/dispatcher/dispatcher.py
JrooTJunior: Since you removed `dispatcher` and `self.workflow_data` from the `workflow_data` dictionary, the startup and shutdown callbacks will no longer receive a Dispatcher instance and other dependencies, so this is incorrect.

PR #1728, opened 2025-10-04
Docs: add guide "Testing your bot" (pytest examples)

stilltryinghard:
This PR adds a new guide explaining how to test bot handlers with pytest.

The guide includes:
- a simple echo handler example
- a callback query example

Both examples use dummy objects to avoid real Telegram API requests.


docs/guide/testing.rst
```diff
@@ -0,0 +1,88 @@
...
+
+   @pytest.mark.asyncio
+   async def test_echo_handler():
+       sent = []
+
+       class DummyMessage:
+           def __init__(self, text):
+               self.text = text
+           async def answer(self, text):
+               sent.append(text)
```
Olegt0rr, line 40: It's better to use special testing tools to create mocks, e.g. MagicMock, AsyncMock.

Olegt0rr (changes requested):
It's a good initiative to add information on how to test application logic. What's confusing is that the example provided is rather primitive and doesn't use standard testing tools.

PR #1770, opened 2026-02-11
Migrate to `uv` build system

andrew000:
# Description

Migrate to `uv` build system

Also:
- Replace Makefile with Justfile
- Bump deps


JrooTJunior:
1. Where can I see a trade-off analysis with a conclusion as to why we need to switch to Justfile? The analysis should include at least the following options: `Makefile`, `Justfile`, `Taskfile`, bash scripts, and possibly a few other known solutions. As the lead developer of this project, I should know why I actually need to install another thing for the same ~sh...~ functionality (I use other tools in other projects for the same purposes, but I know why I actually use them).
2. It is also necessary to add a strong motivation for the reasons for migrating to the new build backend

andrew000:
> 1. Where can I see a trade-off analysis with a conclusion as to why we need to switch to Justfile? The analysis should include at least the following options: `Makefile`, `Justfile`, `Taskfile`, bash scripts, and possibly a few other known solutions. As the lead developer of this project, I should know why I actually need to install another thing for the same ~sh...~ functionality (I use other tools in other projects for the same purposes, but I know why I actually use them).
> 
>     2. It is also necessary to add a strong motivation for the reasons for migrating to the new build backend

Old `Makefile` was Unix-only and created friction for Windows contributors. Looking at the old file, it used Unix-specific commands like `find` and `rm -rf` that don't work on Windows.

`Makefile` is great on Unix, but terrible on Windows, tab-sensitive syntax is annoying, scary error messages, especially with complex command chains... `Just` is native cross-platform, has clean syntax, built-in attributes like `[unix]` to separate recipes running on different OS for example, clear error messages, etc.

Since Aiogram has Windows contributors, `Just` was the obvious choice. The new `Justfile` has explicit `[windows]` and `[unix]` sections that handle platform differences cleanly.

What about `uv_build` over `hatchling` - `aiogram` already uses `uv` as project management tool, so why not to use uv's build backend? One tool for everything - deps and building. `uv` is faster in most operations and saves time in GitHub CI.

It would be cool to gradually rebuild the entire Aiogram ecosystem around `uv`.

<details>
<summary>Also</summary>

#### Why not Taskfile
```
# Taskfile.yml - verbose YAML
tasks:
  lint:
    cmds:
      - uv run ruff format --check --diff {{.package_dir}}
      - uv run ruff check --show-fixes --preview {{.package_dir}} {{.examples_dir}}
      - uv run mypy {{.package_dir}}

# Justfile - cleaner
lint:
    uv run ruff format --check --diff {{ package_dir }}
    uv run ruff check --show-fixes --preview {{ package_dir }} {{ examples_dir }}
    uv run mypy {{ package_dir }}
```
#### Why not "bash scripts"
Bash scripts have cross-platform issues - they simply don't work on Windows without WSL/Cygwin.
</details>

PR #1777, opened 2026-02-27
Add MediaGroupAggregatorMiddleware

### LINKED ISSUE #1697: Add media group message type handler
### aiogram version

3.x

### Problem

When I send N photos (a media group) to the bot, it calls message handler N times once for each attached photo. I think there should be a single call to the handler. Working around with some local dictionaries is bug prone because you have to detect when it is the last image in the bundle and clear dictionary entry to allow user to forward the same message to the bot and handle it again

### Possible solution

Add a decorator for these type of messages that overrides default
```
@dp.message()
```

### Alternatives

I wanted to try `aiogram-mediagroup-handle` but it doesn't install

### Code example

```python3

```

### Additional information

_No response_

Vitaly312:
# Description
Fixes #1697 

Adds `MediaGroupAggregatorMiddleware`  that collects media group data into a single list.
There are both in-memory and Redis implementations.

The code uses a lock and delay approach because Telegram API does not provide any `media_group_finished` event.
Middleware chooses only one message for processing via a locking mechanism; and after some delay calls the handler to process the collected album. Redis aggregator is ready for multi-instance processing with distributed locking and correctly handles retries on errors (favors at-least-once delivery, so in some cases like lock timeout expiration one album may be processed more than once)

## Usage
This middleware should be registered like
```python
from aiogram.dispatcher.middlewares.media_group import MediaGroupAggregatorMiddleware
...

dispatcher = Dispatcher(media_group_aggregator=MediaGroupAggregatorMiddleware())
```
And handlers can use it in this way:
```python
@dispatcher.message(F.media_group_id)
async def start(message: types.Message, album: list[types.Message]):
    print(message.media_group_id)
    print(len(album), [m.message_id for m in album])
```

## Type of change

# How Has This Been Tested?
Unit and integration tests for:
- [x] isolation for parallel requests
- [x] retries if handler failed
- [x] lock managment

**Test Configuration**:
* Operating System: Ubuntu 24.04
* Python version: 3.12

# Checklist:

- [x] I have added tests that prove my feature works


aiogram/filters/media_group.py
```diff
@@ -0,0 +1,62 @@
+from typing import Any, Literal
+
+from aiogram.filters.base import Filter
+from aiogram.types import Message
+
+MIN_MEDIA_COUNT = 2
+DEFAULT_MAX_MEDIA_COUNT = 10
+
+
+class MediaGroupFilter(Filter):
```
JrooTJunior: It is better to change to MagicFilter and document it.

There is already implemented a way to check media count like this:
```python
MagicData(F.media.len() < 3)
```

Vitaly312:
Done. I've implemented the suggestions from Copilot. Now it is ready for human review.

Vitaly312:
@JrooTJunior hi, could you trigger the CI? I’ve resolved suggestions from Copilot.

aiogram/dispatcher/middlewares/media_group.py
```diff
@@ -0,0 +1,246 @@
...
+                        return None
+                    album = sorted(
+                        (msg.as_(cast(Bot, data.get("bot"))) for msg in album),
+                        key=lambda msg: msg.message_id,
+                    )
+                    data.update(album=album)
+                    result = await handler(album[0], data)
+                    await self.media_group_aggregator.delete_group(event.media_group_id)
+                    return result
+                await asyncio.sleep(delta)
```
JrooTJunior, line 244: What happens if a bot has more than 1 replica?
Different instances may receive different media group messages, so which instance will handle this?

docs/utils/media_group.rst
```diff
@@ -1,8 +1,13 @@
 ===================
-Media group builder
+Media group
 ===================
```
JrooTJunior: The Markdown format was changed incorrectly. The title bar should contain exactly the same number of characters as the title.

aiogram/dispatcher/middlewares/media_group.py
Vitaly312: Multi-instance bots must use `RedisMediaGroupAggregator`.

So, in this case each message will be stored in Redis. Each instance tries to acquire the lock, and one of them (usually the one that receives first media, but if processing failed the one that receive a retry from Telegram) will process all messages from Redis.

Vitaly312:
@JrooTJunior thanks for the review. I resolved all requested changes.

Latand:
I ran a code review with Codex and found two behavior issues that look worth addressing before merge.

### 1. Media group aggregation depends on concurrent update processing

The middleware waits inside the first media update task. If updates are processed sequentially, the next album item cannot enter `add_into_group()` until the first update has already slept, emitted a one-item album, and deleted the group.

This is not just a direct middleware-call issue. I reproduced it through the real polling path on PR head `28626d12`:

```python
from datetime import datetime
from typing import Any
from unittest.mock import patch

from aiogram import Dispatcher
from aiogram.dispatcher.middlewares.media_group import MediaGroupAggregatorMiddleware
from aiogram.types import Chat, Message, Update


async def test_media_group_not_aggregated_when_polling_is_sequential(bot):
    dispatcher = Dispatcher()
    dispatcher.message.outer_middleware(MediaGroupAggregatorMiddleware(delay=0.01))
    calls: list[list[int]] = []

    @dispatcher.message()
    async def handler(message: Message, album: list[Message]) -> None:
        calls.append([msg.message_id for msg in album])

    def make_update(update_id: int, message_id: int) -> Update:
        return Update(
            update_id=update_id,
            message=Message(
                message_id=message_id,
                date=datetime.now(),
                chat=Chat(id=1, type="private"),
                media_group_id="42",
            ),
        )

    async def mock_updates(*_: Any):
        yield make_update(100, 1)
        yield make_update(101, 2)

    with patch("aiogram.dispatcher.dispatcher.Dispatcher._listen_updates") as patched:
        patched.return_value = mock_updates()
        await dispatcher._polling(bot=bot, handle_as_tasks=False)

    assert calls == [[1, 2]]
```

Actual result on this PR:

```python
[[1], [2]]
```

I also checked the concurrency-limited case:

```text
handle_as_tasks=False: [[1], [2]]
handle_as_tasks=True, tasks_concurrency_limit=1: [[1], [2]]
handle_as_tasks=True, tasks_concurrency_limit=2: [[1, 2]]
```

So the feature currently works only when updates from the same album can be processed concurrently. I think this either needs to be documented/enforced as a hard requirement, or the aggregation design needs to move out of the blocking update task. A fuller fix would likely require a per-media-group background coordinator/task that accepts messages quickly and invokes the handler after the quiet period, but that has to be designed carefully around error propagation, cancellation, webhook response semantics, and handler context.

### 2. Propagated `message` can disagree with `event_update`

The middleware sorts the album and calls the handler with `album[0]`, but the contextual `event_update` remains the update that acquired the lock. If message 2 arrives first and message 1 arrives second, the handler receives `message_id == 1`, while `event_update.message.message_id == 2`.

Repro:

```python
import asyncio
from datetime import datetime
from typing import Any

from aiogram import Dispatcher
from aiogram.dispatcher.middlewares.media_group import MediaGroupAggregatorMiddleware
from aiogram.types import Chat, Message, Update


async def test_album_event_update_matches_propagated_message(bot):
    dispatcher = Dispatcher()
    dispatcher.message.outer_middleware(MediaGroupAggregatorMiddleware(delay=0.01))
    seen: dict[str, Any] = {}

    @dispatcher.message()
    async def handler(message: Message, event_update: Update, album: list[Message]) -> None:
        seen["message_id"] = message.message_id
        seen["event_update_message_id"] = event_update.message.message_id
        seen["album_ids"] = [msg.message_id for msg in album]

    chat = Chat(id=1, type="private")

    def make_update(update_id: int, message_id: int) -> Update:
        return Update(
            update_id=update_id,
            message=Message(
                message_id=message_id,
                date=datetime.now(),
                chat=chat,
                media_group_id="42",
            ),
        )

    await asyncio.gather(
        dispatcher.feed_update(bot, make_update(100, 2)),
        dispatcher.feed_update(bot, make_update(101, 1)),
    )

    assert seen == {
        "message_id": 1,
        "event_update_message_id": 1,
        "album_ids": [1, 2],
    }
```

Actual result on this PR:

```python
{
    "message_id": 1,
    "event_update_message_id": 2,
    "album_ids": [1, 2],
}
```

This breaks the usual aiogram expectation that injected contextual data describes the same event object passed to the handler. Possible fixes are either to preserve the original event as the handler argument instead of replacing it with `album[0]`, or to rebuild/replace `event_update` and related contextual data consistently when choosing a different propagated message.

aiogram/dispatcher/middlewares/media_group.py
```diff
@@ -0,0 +1,246 @@
...
+                    album = await self.media_group_aggregator.get_group(event.media_group_id)
+                    if not album:
+                        return None
+                    album = sorted(
+                        (msg.as_(cast(Bot, data.get("bot"))) for msg in album),
+                        key=lambda msg: msg.message_id,
+                    )
+                    data.update(album=album)
+                    result = await handler(album[0], data)
+                    await self.media_group_aggregator.delete_group(event.media_group_id)
```
RootShinobi: + memory leak on error in handler

RootShinobi:
maybe https://github.com/RootShinobi/aiogram_album/blob/main/aiogram_album/album_message.py some sugar?

aiogram/dispatcher/middlewares/media_group.py
Vitaly312: On error in the handler media group data will be deleted via TTL in MediaGroupAggregator(both Redis and memory implementations have this). Deleting data immediately after an error happens in handler is unacceptable because it leads to incorrect telegram retry handling.

Vitaly312:
@Latand I've resolved these issues.
  > ### 1. Media group aggregation depends on concurrent update processing
I chose this approach (keeping one media request until the whole group is processed) because it is the only way to handle telegram retries in cases when the handler raises an error and ensure reliability. If the middleware passes the group into some background task and returns immediately, the query's answer will be 200 (if the bot uses webhook) and Telegram will not send any retries. So, now I've described this requirement in the documentation. 

> ### 2. Propagated message can disagree with event_update
Fixed. Now the middleware reconstructs the `event_update` copy and passes it instead of the original object. 

@RootShinobi thanks for the suggestion! I think it would be great to have these features, but it is more convenient to implement them as a separate follow-up PR, after core logic is merged.

Vitaly312:
@JrooTJunior I've implemented all suggestions from the community. Could you approve the CI and review it again?

Vitaly312:
@JrooTJunior I've synced this branch with the latest changes from the dev-3.x branch. This PR is completely ready now. Could you please trigger CI and take another look?

aiogram/dispatcher/middlewares/media_group.py
```diff
@@ -0,0 +1,250 @@
...
+
+    async def __call__(
+        self,
+        handler: Callable[[TelegramObject, dict[str, Any]], Awaitable[Any]],
+        event: TelegramObject,
+        data: dict[str, Any],
+    ) -> Any:
+        if not isinstance(event, Message) or not event.media_group_id:
+            return await handler(event, data)
+        await self.media_group_aggregator.add_into_group(event.media_group_id, event)
```
JrooTJunior: Media groups break when FSM event isolation is enabled.

It sleeps before calling the downstream handler, but message middlewares run inside the update-level FSM middleware. With `SimpleEventIsolation`/`Redis` isolation, the first media update holds the FSM isolation lock while waiting, so the second media update cannot enter the aggregator until after the first is processed.


---

To enable this feature, pass the `events_isolation` argument to the dispatcher:

```python
from aiogram.fsm.storage.memory import SimpleEventIsolation

dispatcher = Dispatcher(events_isolation=SimpleEventIsolation())
```

aiogram/dispatcher/middlewares/media_group.py
```diff
@@ -0,0 +1,250 @@
...
+                    album = await self.media_group_aggregator.get_group(event.media_group_id)
+                    if not album:
+                        return None
+                    album = sorted(
+                        (msg.as_(cast(Bot, data.get("bot"))) for msg in album),
+                        key=lambda msg: msg.message_id,
+                    )
+                    if event_update := cast(Update, data.get("event_update")):
+                        data.update(
+                            event_update=event_update.model_copy(update={"message": album[0]})
```
JrooTJunior: `event_update` is rewritten as message for every Message observer.

If the middleware is attached to `channel_post`, `edited_message`, `business_message`, etc., those events are also `Message` objects and may have `media_group_id`. The copied update keeps the original field and adds message, so `event_update.event_type` becomes message and downstream code sees the wrong update kind.

aiogram/dispatcher/middlewares/media_group.py
```diff
@@ -0,0 +1,250 @@
...
+            lock from expiring before the handler finishes, but small enough to expire before
+            Telegram retries a failed delivery.
+        """
+        self.redis = redis
+        self.ttl_sec = ttl_sec
+        self.lock_ttl_sec = lock_ttl_sec
+
+    @staticmethod
+    def get_group_key(media_group_id: str) -> str:
+        return f"media_group:{media_group_id}:album"
```
JrooTJunior: In shared Redis or multi-bot deployments, this should include at least bot id and probably chat/update type. Otherwise unrelated albums can collide if Telegram’s media_group_id is not globally unique across all bots and chats. For a “distributed environment” storage, the namespace needs to be defensively scoped.

Key builder can helps you to solve this issue: 
https://github.com/aiogram/aiogram/blob/dev-3.x/aiogram/fsm/storage/base.py#L24-L99

aiogram/dispatcher/middlewares/media_group.py
```diff
@@ -0,0 +1,250 @@
...
+        await cast(
+            Awaitable[int],
+            self.redis.eval(release_script, 1, self.get_group_lock_key(media_group_id), lock_id),
+        )
+
+    async def get_group(self, media_group_id: str) -> list[Message]:
+        result = await cast(
+            Awaitable[list[str]], self.redis.lrange(self.get_group_key(media_group_id), 0, -1)
+        )
+        return self.deduplicate_messages([Message.model_validate_json(msg) for msg in result])
```
JrooTJunior: Shortcuts (e.g. `await message.answer(...)`) will not work because the new message object has not been attached to any bot instance.

Possible solution:
https://github.com/aiogram/aiogram/blob/dev-3.x/aiogram/client/session/base.py#L97

JrooTJunior (changes requested):

Peppymiller:
I'm getting an error on the latest commit, bfb645d:

RuntimeError: MediaGroupAggregatorMiddleware got an unexpected event type!

This error does not occur in commit 032bf01

Code example:
https://pastebin.com/U0Yf2bcJ

Vitaly312:
@Peppymiller the actual way to register the middleware is `dispatcher = Dispatcher(media_group_aggregator=MediaGroupAggregatorMiddleware())` instead of `router.message.outer_middleware( MediaGroupAggregatorMiddleware())` (the middleware will be applied to all kinds of events, e.g. message, edited_message, channel_post, etc.)

I've changed this in the commit [a77f288](https://github.com/aiogram/aiogram/pull/1777/commits/a77f288157002490463639f9f763d72c929af8b8) to avoid media group breaking when FSM event isolation is enabled as was mentioned in the https://github.com/aiogram/aiogram/pull/1777#discussion_r3365462088 by registering the middleware before the FSM middleware holds a lock.

Vitaly312:
@JrooTJunior I’ve addressed your feedback. Could you please trigger
Cl and take another look?

PR #1813, opened 2026-05-29
Add Justfile task runner

andrew000:
# Description

Adds a `Justfile` that mirrors the existing `Makefile` development commands while allowing platform-specific implementations where needed.

The goal is to provide a clearer cross-platform command runner for common contributor workflows such as linting, formatting, testing, docs, code generation, and release helpers.

## Motivation

The current `Makefile` is mostly used as a task runner rather than as an incremental build system. Several recipes rely on Unix shell behavior (`rm`, `find`, `mkdir -p`, `$(shell pwd)`), which makes the documented workflow less native for Windows contributors.
 
`just` is designed for this kind of project automation:

- explicit task-runner semantics;
- `just --list` command discoverability;
- native `[unix]` / `[windows]` recipe variants;
- simpler argument forwarding for recipes like `bump` and `docs-serve`;
- same high-level command surface as the existing `Makefile`.

### DIFF
--- a/CHANGES/1813.misc.rst
+++ b/CHANGES/1813.misc.rst
@@ -1,0 +1,1 @@
+Added a ``Justfile`` with project automation recipes mirroring the existing ``Makefile`` workflow.
--- a/Justfile
+++ b/Justfile
@@ -1,0 +1,146 @@
+#!/usr/bin/env just --justfile
+
+set shell := ["bash", "-c"]
+set windows-shell := ["powershell.exe", "-NoLogo", "-Command"]
+
+default: lint
+
+package_dir := "aiogram"
+tests_dir := "tests"
+scripts_dir := "scripts"
+examples_dir := "examples"
+code_dir := f"{{package_dir}} {{tests_dir}} {{scripts_dir}} {{examples_dir}}"
+reports_dir := "reports"
+redis_connection := "redis://localhost:6379"
+mongo_connection := "mongodb://mongo:mongo@localhost:27017"
+
+# Documentation variables
+
+locales := "uk_UA"
+locale_targets := "docs-serve-" + locales
+locales_pot := "_build/gettext"
+docs_dir := "docs"
+
+# =================================================================================================
+# Environment
+# =================================================================================================
+
+[unix]
+clean:
+    find . -name "__pycache__" -type d -exec rm -rf {} +
+    find . -type f -name "*.py[co]" -delete
+    find . -type f -name "*~" -delete
+    find . -type f -name ".*~" -delete
+    find . -name ".pytest_cache" -type d -exec rm -rf {} +
+    rm -rf *.egg-info
+    rm -f report.html
+    rm -f .coverage
+    rm -rf build dist site .cache .mypy_cache .ruff_cache {{ reports_dir }}
+
+[windows]
+clean:
+    -Get-ChildItem -Path . -Recurse -Filter "__pycache__" -Directory | Remove-Item -Recurse -Force
+    -Get-ChildItem -Path . -Recurse -Include "*.pyc", "*.pyo" -File | Remove-Item -Force
+    -Get-ChildItem -Path . -Recurse -Include "*~", ".*~" -File | Remove-Item -Force
+    -Get-ChildItem -Path . -Recurse -Filter ".pytest_cache" -Directory | Remove-Item -Recurse -Force
+    -Remove-Item -Path "*.egg-info" -Recurse -Force -ErrorAction SilentlyContinue
+    -Remove-Item -Path "report.html" -Force -ErrorAction SilentlyContinue
+    -Remove-Item -Path ".coverage" -Force -ErrorAction SilentlyContinue
+    -Remove-Item -Path "build", "dist", "site", ".cache", ".mypy_cache", ".ruff_cache", "{{ reports_dir }}" -Recurse -Force -ErrorAction SilentlyContinue
+
+install: clean
+    uv sync --all-extras --group dev --group test
+    uv run pre-commit install
+
+# =================================================================================================
+# Code quality
+# =================================================================================================
+
+[default]
+lint:
+    uv run ruff format --check --diff {{ package_dir }}
+    uv run ruff check --show-fixes --preview {{ package_dir }} {{ examples_dir }}
+    uv run mypy --native-parser --num-workers 8 {{ package_dir }}
+
+reformat:
+    uv run ruff format {{ code_dir }}
+    uv run ruff check --fix {{ code_dir }}
+
+# =================================================================================================
+# Tests
+# =================================================================================================
+
+test-run-services:
+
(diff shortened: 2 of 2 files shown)


andrew000:
# Justfile Migration

## Options considered

### 1. Keep `Makefile`

Pros:

- Already exists and works for maintainers familiar with Unix tooling.
- No new dependency for Linux/macOS contributors.
- `make` is universal in many open source projects.
- Simple recipes like `make lint`, `make test`, and `make build` are already understandable.

Cons:

- Poor native Windows support.
- The current `Makefile` uses Unix shell commands such as `rm`, `find`, backticks, `mkdir -p`, and `$(shell pwd)`.
- Windows contributors need Git Bash, WSL, MSYS2, or a similar compatibility layer.
- Make syntax is powerful but awkward for task-runner use cases:
  - argument forwarding is clumsy;
  - quoting rules are surprising;
  - recipes are tied to shell behavior;
  - `.PHONY` and Make-specific semantics add noise for simple commands.
- Make is primarily a build dependency graph tool, while this repository mostly uses it as a command alias/task runner.

Conclusion:

Keeping only `Makefile` is the lowest-risk option if we do not care about native Windows ergonomics. It is probably sufficient for maintainers who already use Unix-like environments. It is less ideal for contributors working directly on Windows.

### 2. Add or migrate to `Justfile`

Pros:

- `just` is designed specifically as a command runner, not as a build system.
- Recipes are easier to read than Make recipes.
- Cross-platform recipes are first-class:
  - `[unix]`
  - `[windows]`
  - configurable shells
- Better developer UX:
  - `just --list` gives discoverable commands;
  - arguments are explicit;
  - default recipes are simple;
  - no `.PHONY` boilerplate.
- Allows us to preserve the same command names:
  - `just lint`
  - `just test`
  - `just build`
  - `just update-api`
- It can provide native Windows equivalents for commands like cleanup and report directory creation without forcing WSL or Git Bash.
- It is small and focused. It does not try to manage Python environments, dependency resolution, or packaging.

Cons:

- It is another tool contributors must install.
- It duplicates the existing `Makefile` unless we fully migrate.
- It is not Python-native.
- If the `Justfile` only mirrors the `Makefile`, it creates maintenance burden without enough value.
- Some contributors may already have `make` installed but not `just`.

Conclusion:

`Justfile` is justified only if the project wants a cross-platform task runner with native Windows support and better command discoverability. If that is not a project goal, then adding it is unnecessary duplication.

### 3. Taskfile / `go-task`

Pros:

- Cross-platform.
- YAML-based, familiar to many users.
- Good support for task dependencies, variables, includes, and descriptions.
- Can replace many Makefile patterns cleanly.

Cons:

- Heavier than `just` for this use case.
- YAML is more verbose for simple command aliases.
- Adds another external binary, the same as `just`.
- More features than this repository appears to need.
- Less common in Python projects than Make or simple task runners.

Conclusion:

Taskfile is a good solution for larger workflows with structured task composition. For aiogram's current needs, it is probably more tool than necessary.

### 4. Bash scripts

Pros:

- No special task-runner syntax.
- Easy for Unix users.
- Can express complex behavior clearly when scripts are well structured.
- Existing Makefile recipes could call scripts, reducing Makefile complexity.

Cons:

- Not natively cross-platform.
- Windows users still need WSL, Git Bash, or MSYS2.
- Discoverability is worse unless we add a wrapper command or documentation.
- Argument handling and command organization can become inconsistent across scripts.
- We would likely end up recreating a small task runner manually.

Conclusion:

Bash scripts are good for complex Unix-specific automation, but they do not solve the Windows support problem. They are not a strong replacement for a project command runner here.

### 5. Python scripts

Pros:

- Python is already required by the project.
- Native cross-platform behavior.
- No extra non-Python binary.
- Good for complex automation where we want testable logic.

Cons:

- More verbose for simple command aliases.
- A custom CLI would need maintenance.
- We would need to decide how to expose commands:
  - `python scripts/dev.py lint`
  - `uv run python scripts/dev.py lint`
  - console script
- Shelling out to `ruff`, `pytest`, `mypy`, etc. from Python adds boilerplate.
- Less familiar than `make lint` / `just lint` for open source contributors.

Conclusion:

Python scripts are the strongest alternative if avoiding another binary is the main priority. However, for simple task orchestration, they are more code and maintenance than `just`.

### 6. `tox` / `nox`

Pros:

- Python-native ecosystem.
- Good for test matrices.
- Good for isolated environments.
- Familiar in Python projects.

Cons:

- This project already uses `uv` directly.
- `tox` / `nox` would overlap with existing dependency and environment management.
- Less convenient for non-test project tasks like:
  - `update-api`
  - `towncrier-draft-github`
  - docs serving
  - release tagging
- Adds configuration complexity for commands that are currently simple.

Conclusion:

Good for environment and test matrix management, but not the best fit as a general replacement for the current Makefile.

## Why `just` would be preferable if we migrate

The strongest reason is native cross-platform contributor workflow.

aiogram supports Python `>=3.10`, including Windows users. The current `Makefile` is effectively Unix-oriented. A contributor on Windows can use WSL or Git Bash, but that means the documented development commands are not equally native across supported platforms.

A `Justfile` lets us keep the same high-level commands while writing platform-specific implementations only where needed:

- Unix cleanup can use `find` and `rm`.
- Windows cleanup can use PowerShell.
- Shared recipes like lint, test, docs, and codegen stay identical.

This makes the development workflow more honest: if the project supports contributors on Windows, the command runner should not assume Unix shell tools for basic maintenance tasks.

The second reason is command-runner semantics.

The repository is not using Make as a real incremental build system. There are no file targets where timestamp-based rebuild logic is important. The file is mostly a list of named commands:

- `clean`
- `install`
- `lint`
- `reformat`
- `test`
- `docs-serve`
- `build`
- `update-api`
- `release`

That is exactly the use case `just` is designed for. It removes Make-specific noise and makes the file easier to treat as contributor-facing automation.

The third reason is discoverability and maintainability.

`just --list` gives a clear list of available project commands. Recipe arguments such as `bump *args` and `docs-serve *OPTS` are also more explicit than Make variable forwarding. That matters for contributors who do not touch release or codegen commands often.

## Proposed conclusion

My conclusion is:

If aiogram wants to keep the existing Unix-first maintainer workflow and does not want another required tool, we should keep the `Makefile` and close this change.

If aiogram wants native Windows-compatible project commands, clearer task-runner semantics, and better command discoverability, then `just` is a reasonable migration target. In that case, the `Justfile` should not be a partial duplicate. It should either become the documented primary command runner, or be introduced temporarily with a clear follow-up plan to remove the `Makefile`.

For this PR specifically, the motivation should be:

- make contributor automation native on Windows, macOS, and Linux;
- replace Make-as-task-runner usage with a tool designed for task running;
- keep the existing command surface mostly unchanged;
- reduce shell-specific assumptions in project commands;
- improve command discoverability via `just --list`.

Without those goals being accepted by maintainers, adding `Justfile` is not worth the extra dependency.

PR #1818, opened 2026-06-01
fix(types): allow missing ChatMemberRestricted can_react_to_messages

redactdeveloper:
Summary
- make ChatMemberRestricted.can_react_to_messages optional when Telegram omits it
- update butcher metadata so regeneration preserves the optional field
- add regression coverage for restricted chat members without can_react_to_messages

Tests
- python -m pytest tests/test_utils/test_chat_member.py
- python -m ruff check aiogram tests
- python -m ruff format --check aiogram tests

Fixes #1812

PR #1824, opened 2026-06-01
fix(types): make can_react_to_messages and can_edit_tag optional in ChatMemberRestricted

### LINKED ISSUE #1812: Pydantic ValidationError on ChatMemberRestricted: missing can_react_to_messages field
### Checklist

### Operating system

Ubuntu24

### Python version

3.12.3

### aiogram version

3.28.2

### Expected behavior

The my_chat_member update with restricted status should be parsed successfully into the ChatMemberRestricted model without throwing any Pydantic ValidationError.

### Current behavior

When a bot (or user) receives a my_chat_member update where the status is changed to restricted, aiogram crashes and raises a ValidationError. It seems the Telegram API sometimes omits the can_react_to_messages field for restricted members, but aiogram strictly requires it as a boolean.

### Steps to reproduce

Running a bot using aiogram 3.28.2.

The bot's permissions in a Telegram group were changed, setting its status to "Restricted."

The update sent by the Telegram API is missing the `can_react_to_messages` field.

The bot's dispatcher is unable to parse the update and raises a `ValidationError` exception.

### Code example

```python3

```

### Logs

```sh
pydantic_core._pydantic_core.ValidationError: 17 validation errors for Response[list[Update]]
result.0.my_chat_member.new_chat_member.ChatMemberRestricted.can_react_to_messages
  Field required [type=missing, input_value={'user': {'id': ...True, 'is_member': True}, input_type=dict]
```

### Additional information

In aiogram/types/chat_member_restricted.py, change:
can_react_to_messages: bool
to:
can_react_to_messages: bool | None = None

I discovered this bug with the help of AI, and the issue was written by the AI.

rusthype:
## Problem

Fixes #1812

When a bot receives a `my_chat_member` update with `status: restricted`, Telegram **sometimes omits** the `can_react_to_messages` and `can_edit_tag` fields. This causes a hard crash:

```
pydantic_core.ValidationError: 2 validation errors for Response[list[Update]]
  ...can_react_to_messages
    Field required [type=missing, ...]
  ...can_edit_tag
    Field required [type=missing, ...]
```

Both fields were added in newer Bot API versions and are not always present in responses for restricted members — particularly when the restriction was applied before these fields were introduced.

## Fix

Mark both fields as `Optional` with a default of `None`:

```python
# Before
can_react_to_messages: bool
can_edit_tag: bool

# After
can_react_to_messages: bool | None = None
can_edit_tag: bool | None = None
```

The `.butcher` schema is also updated so re-running codegen does not revert this fix.

## Precedent

This mirrors the pattern already used for `tag: str | None = None` in the same model, and matches how `ChatPermissions` handles its optional boolean fields.

## Tests

Added `tests/test_api/test_types/test_chat_member_restricted.py` covering:
- Both fields present (existing behaviour)
- `can_react_to_messages` absent → parses cleanly, value is `None`
- `can_edit_tag` absent → parses cleanly, value is `None`
- Both absent simultaneously (the real-world scenario from #1812)

### DIFF
--- a/.butcher/types/ChatMemberRestricted/entity.json
+++ b/.butcher/types/ChatMemberRestricted/entity.json
@@ -126,18 +126,18 @@
       {
         "type": "Boolean",
         "description": "True, if the user is allowed to react to messages",
-        "html_description": "<td><em>True</em>, if the user is allowed to react to messages</td>",
-        "rst_description": ":code:`True`, if the user is allowed to react to messages\n",
+        "html_description": "<td><em>Optional</em>. <em>True</em>, if the user is allowed to react to messages</td>",
+        "rst_description": "*Optional*. :code:`True`, if the user is allowed to react to messages\n",
         "name": "can_react_to_messages",
-        "required": true
+        "required": false
       },
       {
         "type": "Boolean",
         "description": "True, if the user is allowed to edit their own tag",
-        "html_description": "<td><em>True</em>, if the user is allowed to edit their own tag</td>",
-        "rst_description": ":code:`True`, if the user is allowed to edit their own tag\n",
+        "html_description": "<td><em>Optional</em>. <em>True</em>, if the user is allowed to edit their own tag</td>",
+        "rst_description": "*Optional*. :code:`True`, if the user is allowed to edit their own tag\n",
         "name": "can_edit_tag",
-        "required": true
+        "required": false
       },
       {
         "type": "Boolean",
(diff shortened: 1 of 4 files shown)


JrooTJunior:
> The .butcher schema is also updated so re-running codegen does not revert this fix.

You are wrong, the Bot API documentation parser will completely undo these changes.
You should never touch the codegen part. The answer is the same as in the same PR: https://github.com/aiogram/aiogram/pull/1818#pullrequestreview-4400227650

PR #1829, opened 2026-06-09
fix(dispatcher): detect async lambdas and callable objects correctly

### LINKED ISSUE #1721: lambdas and callable objects are not recognized as awaitable handlers

byArto:
## Summary

Detect async handlers implemented as callable objects with async __call__ method or lambdas that return coroutines (e.g., lambda m: async_func(m)).

## Problem

Previously only plain async functions and classes extending BaseHandler were detected as awaitable. This caused coroutine was never awaited warnings for callable objects with async __call__ and lambdas wrapping async functions.

## Changes

- CallableObject.__post_init__() now checks __call__ method for callable objects
- Added _check_awaitable_via_call() helper that tries calling the callback with a dummy argument to detect if the result is awaitable
- Added test coverage for AsyncCallable class and lambda returning coroutine

Closes #1721

Co-authored-by: Claude <claude@anthropic.com>


JrooTJunior:
Are you kidding?

image

Please rebase this PR to `dev-3.x` branch

PR #1830, opened 2026-06-11
Added support of the Bot API 10.1

JrooTJunior:
Added full support of the [Bot API 10.1](https://core.telegram.org/bots/api-changelog#june-11-2026)


aiogram/types/message.py
```diff
@@ -3979,11 +3989,12 @@ def copy_to(
 
     def edit_text(
         self,
-        text: str,
         inline_message_id: str | None = None,
+        text: str | None = None,
```
Olegt0rr, line 3992: It would be nice to pin `text` first, to prevent bugs with positional args

```python
await message.edit_text("new text")
```
 
This will bind `"new text"` to `inline_message_id`, leaving `text=None`. There's no error at call time – you get either wrong behavior or a runtime error from Telegram ("text or rich_message is required"). This will silently break a large number of existing bots on upgrade.

OR

if we want to really break it – it better to use `*` guard

Olegt0rr (changes requested):

aiogram/types/message.py
JrooTJunior: Exactly what I was saying in the core-dev chat, you're right, this is a critical change, especially considering it's a popular API method.

I would prefer to add `*` everywhere, but that's too big a change for a minor update, so I try to adapt the transformation step in the code generation process to preserve order where possible.

KiGamji: huh, was it a server-side change?

JrooTJunior: On the server side, this argument has been changed from mandatory to optional.

It is located in the same position in the documentation as before (after `inline_message_id, before `parse_mode`), but it was previously marked as Required, and has now been changed to Optional.

In Python, required arguments should be placed at the beginning, and optional (or with default values) ones later, so during code generation I sort all arguments by the "required" flag and keep the other arguments in the same position as specified in the documentation, but now when `text` becomes optional, it moves to where it should have always been.

image

So, I'm working on new logic in the transformation step to keep such arguments in the same order as before (except for cases where some of the required arguments remain unchanged).

JrooTJunior: All changes have been regenerated. So, from now on, all new arguments will be added to the end (if possible).

PR #1857, opened 2026-07-20
Fix live photo identification

### LINKED ISSUE #1841: message.content_type returns photo instead of live_photo for Live Photos
### Checklist

### Operating system

ubuntu 24 server

### Python version

3.12

### aiogram version

3.29

### Expected behavior

If a message contains a live_photo, then:

message.content_type == ContentType.LIVE_PHOTO

even though message.photo is also populated.

### Current behavior

message.content_type == ContentType.PHOTO

because message.photo exists.

### Steps to reproduce

1-Create a bot using aiogram.
2-Register a handler that prints the message content type.
3-Send a Live Photo to the bot from Telegram.
4-Observe that message.content_type is photo instead of live_photo.

### Code example

```python3
from aiogram import Bot, Dispatcher, F 
from aiogram.types import Message

dp = Dispatcher()

@dp.message()
async def handler(message: Message): 
     print("content_type:", message.content_type) 
     print("photo:", message.photo is not None)
     print("live_photo:", message.live_photo is not None)
```

### Logs

```sh

```

### Additional information

Send a Live Photo to the bot.

Output:

content_type: photo
photo: True
live_photo: True

Expected output:

content_type: live_photo
photo: True
live_photo: True

When receiving a Telegram Live Photo, message.content_type returns ContentType.PHOTO instead of ContentType.LIVE_PHOTO.

This appears to happen because Live Photos include a regular photo object in the message payload in addition to the live_photo field. The current detection logic checks for message.photo before message.live_photo, causing the content type to be identified as photo.

I believe ContentType.LIVE_PHOTO should take precedence whenever message.live_photo is present.

Moving the live_photo check above the photo (from line 882-883 to line 684-685) check in aiogram/types/message.py resolves the issue.

OldBuddy:
# Description

* fix: invalid `content_type` property identification for a live photo message.
* refactor: convert `if` blocks to `match` statement on `content_type` property.

Fixes #1841 

## Type of change

# Checklist:

### DIFF
--- a/aiogram/types/message.py
+++ b/aiogram/types/message.py
@@ -40,6 +40,7 @@ if TYPE_CHECKING:
         SendDocument,
         SendGame,
         SendInvoice,
+        SendLivePhoto,
         SendLocation,
         SendMediaGroup,
         SendMessage,
@@ -712,6 +713,8 @@ class Message(MaybeInaccessibleMessage):
             return ContentType.DOCUMENT
         if self.game:
             return ContentType.GAME
+        if self.live_photo:
+            return ContentType.LIVE_PHOTO
         if self.photo:
             return ContentType.PHOTO
         if self.sticker:
@@ -850,8 +853,6 @@ class Message(MaybeInaccessibleMessage):
             return ContentType.POLL_OPTION_ADDED
         if self.poll_option_deleted:
             return ContentType.POLL_OPTION_DELETED
-        if self.live_photo:
-            return ContentType.LIVE_PHOTO
         if self.rich_message:
             return ContentType.RICH_MESSAGE
         if self.community_chat_added:
@@ -3930,6 +3931,7 @@ class Message(MaybeInaccessibleMessage):
         | SendDocument
         | SendLocation
         | SendMessage
+        | SendLivePhoto
         | SendPhoto
         | SendPoll
         | SendDice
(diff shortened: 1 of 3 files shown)


aiogram/types/message.py
```diff
@@ -3998,118 +4001,127 @@ def send_copy(  # noqa: C901
...
-            ).as_(self._bot)
-        if self.story:
-            return ForwardMessage(
-                from_chat_id=self.chat.id,
-                message_id=self.message_id,
-                **kwargs,
-            ).as_(self._bot)
-
-        raise TypeError("This type of message can't be copied.")
+        match (self.content_type):
```
JrooTJunior, line 4004: These changes are not related to the original problem, so they should not be changed here.

Another point is that this change reduces performance because instead of directly checking the attributes, it determines the content type and then falls back to the logic.

At the same time, when directly checking attributes, it tells type checkers that the checked attribute is not None, so it unwraps the "Optional" layer from the type, and when the attribute is passed to an API method, it clearly remains as required/passed, but when you don't check, it remains optional and is incompatible with API calls where these attributes remain required.

JrooTJunior (changes requested):

JrooTJunior:
since no response or changes were received, this was raised in another PR

PR #1862, opened 2026-08-01
Fix Markdown expandable_blockquote missing the expandability mark

uttam12331:
### Problem

`MarkdownDecoration.expandable_blockquote` produces the same output as a plain blockquote with a bare `||` appended — it never emits the `**>` that MarkdownV2 uses to mark a block quotation as *expandable*:

```python
from aiogram.utils.text_decorations import markdown_decoration as md

md.expandable_blockquote("secret")   # -> '>secret||'
md.blockquote("secret")              # -> '>secret'
```

Per the [MarkdownV2 style](https://core.telegram.org/bots/api#markdownv2-style), an expandable block quotation must prefix the **first** line with `**>` and end the last line with `||`. Without the `**>`, a lone trailing `||` is not the expandability mark — Telegram reads it as the start of a spoiler (`||...||`) that is never closed, so sending the result with `parse_mode="MarkdownV2"` fails with `can't parse entities: Can't find end of Spoiler entity...`. At best, the expandable property is silently lost.

This also affects `aiogram.utils.formatting.Text(...).as_markdown()`, which routes an `ExpandableBlockquote` node through this method.

### Cause

```python
def expandable_blockquote(self, value: str) -> str:
    return "\n".join(f">{line}" for line in value.splitlines()) + "||"
```

The HTML sibling `HtmlDecoration.expandable_blockquote` correctly preserves expandability (`<blockquote expandable>`); the Markdown one drops it.

### Fix

Prefix the first line with `**>` (keeping `>` for the remaining lines):

```python
def expandable_blockquote(self, value: str) -> str:
    return (
        "\n".join(
            f"**>{line}" if index == 0 else f">{line}"
            for index, line in enumerate(value.splitlines())
        )
        + "||"
    )
```

Now `expandable_blockquote("secret")` → `**>secret||` and `expandable_blockquote("l1\nl2")` → `**>l1\n>l2||`.

### Tests

Updated the existing `expandable_blockquote` case in `test_text_decorations.py` (which asserted the old `>test||`) to `**>test||`, and added a multi-line test asserting only the first line carries the `**>` mark. Both fail on `dev-3.x` and pass with the fix; the rest of the module is unaffected.

### DIFF
--- a/aiogram/utils/text_decorations.py
+++ b/aiogram/utils/text_decorations.py
@@ -330,7 +330,13 @@ class MarkdownDecoration(TextDecoration):
         return "\n".join(f">{line}" for line in value.splitlines())
 
     def expandable_blockquote(self, value: str) -> str:
-        return "\n".join(f">{line}" for line in value.splitlines()) + "||"
+        return (
+            "\n".join(
+                f"**>{line}" if index == 0 else f">{line}"
+                for index, line in enumerate(value.splitlines())
+            )
+            + "||"
+        )
 
     def date_time(
         self,
--- a/tests/test_utils/test_text_decorations.py
+++ b/tests/test_utils/test_text_decorations.py
@@ -135,7 +135,7 @@ class TestTextDecoration:
             [
                 markdown_decoration,
                 MessageEntity(type="expandable_blockquote", offset=0, length=5),
-                ">test||",
+                "**>test||",
             ],
             [
                 markdown_decoration,
@@ -164,6 +164,13 @@ class TestTextDecoration:
         self, decorator: TextDecoration, entity: MessageEntity, result: str
     ):
         assert decorator.apply_entity(entity, "test") == result
+
+    def test_markdown_expandable_blockquote_multiline(self):
+        # Only the first line carries the ``**>`` expandability mark; the
+        # remaining lines use ``>`` like a normal blockquote, and the closing
+        # ``||`` follows the last line.
+        result = markdown_decoration.expandable_blockquote("line1\nline2\nline3")
+        assert result == "**>line1\n>line2\n>line3||"
 
     @pytest.mark.parametrize(
         "decorator,date_time_format,expected",
--- a/CHANGES/1862.bugfix.rst
+++ b/CHANGES/1862.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed :code:`MarkdownDecoration.expandable_blockquote` not emitting the ``**>`` expandability mark, so it produced invalid MarkdownV2 (a lone trailing ``||``) that Telegram rejected instead of a valid expandable block quotation.


aiogram/utils/text_decorations.py
```diff
@@ -330,7 +330,13 @@ def blockquote(self, value: str) -> str:
...
 
     def expandable_blockquote(self, value: str) -> str:
-        return "\n".join(f">{line}" for line in value.splitlines()) + "||"
+        return (
+            "\n".join(
+                f"**>{line}" if index == 0 else f">{line}"
+                for index, line in enumerate(value.splitlines())
+            )
+            + "||"
+        )
```
JrooTJunior, line 339: `blockquote()` right above already does the per-line prefixing, so this can reuse it:

```suggestion
        return "**" + self.blockquote(value) + "||"
```

Byte-identical output for every non-empty input (checked `'a'`, `'a\nb'`, `'a\n\nb'`, `'\na'`, `'a\n'`), and it keeps the two blockquote methods from drifting apart if line handling ever changes.

JrooTJunior, line 339: Not introduced here, but worth knowing: `expandable_blockquote("")` returns `"||"` — the same unclosed-spoiler failure this PR fixes, reached via a different door. Not reachable from real `MessageEntity` objects (`length >= 1`), but `ExpandableBlockQuote().as_markdown()` gets there. Fine to leave for a follow-up.

tests/test_utils/test_text_decorations.py
```diff
@@ -165,6 +165,13 @@ def test_apply_single_entity(
     ):
         assert decorator.apply_entity(entity, "test") == result
 
+    def test_markdown_expandable_blockquote_multiline(self):
+        # Only the first line carries the ``**>`` expandability mark; the
+        # remaining lines use ``>`` like a normal blockquote, and the closing
+        # ``||`` follows the last line.
```
JrooTJunior, line 171: The comment is longer than the test. `assert result == "**>line1\n>line2\n>line3||"` already says all three things.

JrooTJunior (approved):
Verified locally on `dev-3.x` + this branch:

- `md.unparse('before\nq1\nq2\nafter', [expandable_blockquote@7:5])` → `'before\n**>q1\n>q2||\nafter'`, matching the MarkdownV2 spec (`**>` first line, `>` after, `||` on the last).
- `pytest tests` — 1861 passed, 51 skipped; `ruff check --preview aiogram examples`, `ruff format --check`, `mypy aiogram` all clean.

Worth stating in the changelog that this also fixed `message.md_text`: any received message containing an expandable blockquote previously rendered markup that `parse_mode="MarkdownV2"` rejects outright, so the impact is wider than the constructor path the description leads with.

Two description-only nits: the class is `ExpandableBlockQuote` (capital Q), and the changelog mixes `:code:` and ``` `` ``` roles in one sentence.

JrooTJunior:
Any news?

Mukller (commented):
Verified locally against the Telegram MarkdownV2 spec (CPython 3.13 / Windows):

- `dev-3.x` today: `expandable_blockquote("line1\nline2")` → `">line1\n>line2||"` — no expandability mark, so this serializes a plain blockquote with a literal trailing `||`, which Telegram rejects as invalid.
- This branch: `"**>line1\n>line2||"` — matches the documented expandable block quotation syntax (`**>` on the first line, `||` terminator), single-line case included (`"**>single||"`).
- `html_decoration.expandable_blockquote` is untouched and still emits `<blockquote expandable>`.
- `pytest tests/test_utils/test_text_decorations.py`: 78 passed on this branch, including the updated round-trip case and the new multiline assertion.

One pre-existing edge, not introduced here but visible in both branches: an empty string yields just `"||"`. Probably harmless since entities are never empty in practice, but if you want it bulletproof a guard could raise or return `""`.

Fix is correct from my side.

PR #1879, opened 2026-08-23
Fix client session middleware type annotations to match the decoded result

### LINKED ISSUE #1723: RequestMiddleware incorrect typing
### Checklist

### Operating system

Artix

### Python version

Python 3.11.11

### aiogram version

3.22.0

### Expected behavior

according to the middleware type hints - return type should be a generic `Response[TelegramType]` containing `ok`, `result`, `description`, `error_code` and `parameters`

### Current behavior

in fact, `make_request` of `AiohttpSession`, which is wrapped in the `__call__` of parent class `BaseSession` returns `response.result`  of some concrete `TelegramType` class. That means that every middleware receives `TelegramType` instead of `Response[TelegramType]` making type-hints incorrect.

### Steps to reproduce

set `TG_BOT_TOKEN` and `TG_USER_ID` environment variables and run MRE

### Code example

```python3
import asyncio
import os
from typing import reveal_type

from aiogram import Bot
from aiogram.client.session.aiohttp import AiohttpSession
from aiogram.client.session.middlewares.base import (
    BaseRequestMiddleware,
    NextRequestMiddlewareType,
)
from aiogram.methods import Response, TelegramMethod
from aiogram.methods.base import TelegramType

BOT_TOKEN = os.environ["TG_BOT_TOKEN"]
USER_ID = int(os.environ["TG_USER_ID"])


class TestMiddleware(BaseRequestMiddleware):
    async def __call__(
        self,
        make_request: NextRequestMiddlewareType[TelegramType],
        bot: Bot,
        method: TelegramMethod[TelegramType],
    ) -> Response[TelegramType]:
        resp = await make_request(bot, method)
        reveal_type(resp)
        return resp


async def main():
    async with AiohttpSession() as session:
        session.middleware.register(TestMiddleware())
        bot = Bot(token=BOT_TOKEN, session=session)
        await bot.send_message(chat_id=USER_ID, text="Test!")


if __name__ == "__main__":
    asyncio.run(main())
```

### Logs

```sh
~/Code/tmp (main) > uv run test.py
Runtime type is 'Message'
```

### Additional information

I'm not sure what's the correct approach to fix it. Two solutions as I see it:
- return proper `Response[TelegramType]` from `BaseSession.check_response`, fix the type-hints in `check_response` and `make_request` and return `response.result` in the `__call__` method of `BaseSession`. Bot still receives concrete type, every middleware gets the `Response[TelegramType]` class instead of `TelegramType`, everybody happy
- fix the type-hints of: `NextRequestMiddlewareType`, `RequestMiddlewareType`, `BaseRequestMiddleware` and `BaseRequestMiddleware` to reflect that the `make_request` is actually returning `TelegramType` instead of `Response[TelegramType]`

Mukller:
﻿### Description

Client session middlewares are annotated to receive and return `Response[TelegramType]`, but the actual request chain works with the decoded Bot API result. The mismatch is papered over by a `cast` in `BaseSession.__call__`, so type checkers make every custom session middleware look wrong.

Closes #1723

### Root cause

The runtime path unwraps the envelope long before the middleware boundary:

1. `AiohttpSession.make_request()` calls `check_response()` and returns the decoded `TelegramType` (the `result` field), not a `Response`.
2. `RequestMiddlewareManager.wrap_middlewares()` wraps that callback into the middleware chain.
3. `BaseSession.__call__()` needs `cast(TelegramType, await middleware(bot, method))` precisely because the protocol claims the chain still returns a raw `Response`.

Meanwhile the annotations in `aiogram/client/session/middlewares/base.py` (`NextRequestMiddlewareType`, `RequestMiddlewareType`, `BaseRequestMiddleware`) still declare `Response[TelegramType]`, and the built-in `RequestLogging` middleware repeats it. The documentation examples in `docs/api/session/middleware.rst` teach the same incorrect signature, so users following the docs get type errors in correct code.

### Changes

- `NextRequestMiddlewareType`, `RequestMiddlewareType`, `BaseRequestMiddleware`: return type `Response[TelegramType]` -> `TelegramType`; docstring updated to describe the decoded result
- `RequestLogging`: annotation aligned with the corrected protocols
- Docs examples updated (both code blocks on the session middlewares page)
- Regression test asserting a middleware sees the decoded sentinel object, not a wrapped response
- CHANGES fragment `1723.bugfix.rst`

Note: `check_response()` in `session/base.py` intentionally keeps using `Response` — it operates on the raw envelope internally before unwrapping; this PR does not touch it.

### Testing

```console
python -m pytest tests/test_api/test_client/test_session/ -q
59 passed
```

Including the new `test_middleware_receives_decoded_result`. No runtime behavior change: annotations, docstrings and docs only, plus the test.

### DIFF
--- a/CHANGES/1723.bugfix.rst
+++ b/CHANGES/1723.bugfix.rst
@@ -1,0 +1,9 @@
+Fixed the type annotations of client session middlewares.
+
+:class:`aiogram.client.session.middlewares.base.BaseRequestMiddleware`,
+:class:`aiogram.client.session.middlewares.base.RequestMiddlewareType` and
+:class:`aiogram.client.session.middlewares.base.NextRequestMiddlewareType`
+declared ``Response[TelegramType]`` as the middleware return type, while the
+actual request chain already works with the decoded Bot API result
+(the ``result`` field). The built-in :class:`~aiogram.client.session.middlewares.request_logging.RequestLogging`
+middleware and the documentation examples were updated accordingly.
--- a/aiogram/client/session/middlewares/base.py
+++ b/aiogram/client/session/middlewares/base.py
@@ -7,7 +7,7 @@ from aiogram.methods.base import TelegramType
 
 if TYPE_CHECKING:
     from aiogram.client.bot import Bot
-    from aiogram.methods import Response, TelegramMethod
+    from aiogram.methods import TelegramMethod
 
 
 class NextRequestMiddlewareType(Protocol[TelegramType]):  # pragma: no cover
@@ -15,7 +15,7 @@ class NextRequestMiddlewareType(Protocol[TelegramType]):  # pragma: no cover
         self,
         bot: Bot,
         method: TelegramMethod[TelegramType],
-    ) -> Response[TelegramType]:
+    ) -> TelegramType:
         pass
 
 
@@ -25,7 +25,7 @@ class RequestMiddlewareType(Protocol):  # pragma: no cover
         make_request: NextRequestMiddlewareType[TelegramType],
         bot: Bot,
         method: TelegramMethod[TelegramType],
-    ) -> Response[TelegramType]:
+    ) -> TelegramType:
         pass
 
 
@@ -40,7 +40,7 @@ class BaseRequestMiddleware(ABC):
         make_request: NextRequestMiddlewareType[TelegramType],
         bot: Bot,
         method: TelegramMethod[TelegramType],
-    ) -> Response[TelegramType]:
+    ) -> TelegramType:
         """
         Execute middleware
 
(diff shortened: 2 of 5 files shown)


Mukller:
﻿The red `changes-required` check is not about this branch - it fails before towncrier even runs. The `Checkout code` step of the changelog workflow errors out on every fork PR now:

```
##[error]Refusing to check out fork pull request code from a 'pull_request_target' workflow.
To opt in ... set 'allow-unsafe-pr-checkout: true' on the actions/checkout step.
```

The CHANGES fragment (`1723.bugfix.rst`) is present in the diff and `towncrier check --compare-with origin/dev-3.x` passes locally with both towncrier 24.8 and 25.8. Same failure is visible on #1878, so it affects all external PRs at the moment.

Filed details separately so maintainers can decide how to fix the workflow.

PR #1895, opened 2026-09-02
fix: drop the self-assignment at the top of ContextInstanceMixin.get_current

darkdi:
`get_current` opens with `cls.__context_instance = cls.__context_instance`. `__init_subclass__` already sets that attribute on every subclass, so the read and the write hit the same entry in the class dict and the line is a no-op. It reads as if it were priming something, which is what sent me looking.

### DIFF
--- a/CHANGES/1895.misc.rst
+++ b/CHANGES/1895.misc.rst
@@ -1,0 +1,2 @@
+Removed a statement in :code:`ContextInstanceMixin.get_current` that assigned the
+context variable attribute to itself and had no effect.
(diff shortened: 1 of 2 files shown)


darkdi:
Fragment is in as CHANGES/1895.misc.rst, but changes-required looks like it cannot go green on any fork PR at the moment: actions/checkout refuses to check out fork code under pull_request_target unless the step sets allow-unsafe-pr-checkout: true, so the job dies before towncrier runs

darkdi:
Following up on the CI blocker above: the changelog fragment is present, but `changes-required` still stops during checkout before towncrier runs. Could a maintainer arrange a trusted CI run for this PR? The contributor branch cannot change how the upstream pull_request_target workflow checks out fork code.

PR #1904, opened 2026-09-06
fix: return False for non-ASCII hash in check_webapp_signature

### LINKED ISSUE #1903: check_webapp_signature raises TypeError on non-ASCII hash instead of returning False
### Checklist

### Operating system

Windows 11

### Python version

3.14.3

### aiogram version

3.x (dev-3.x)

### Expected behavior

check_webapp_signature should return False for invalid client-supplied hash values, including hashes that contain non-ASCII characters. The function is annotated -> bool, and the other invalid-input paths in the same function already return False.

### Current behavior

When the hash contains non-ASCII characters (for example after percent-decoding %C3%A9 to "é"), the function raises:

TypeError: comparing strings with non-ASCII characters is not supported

This comes from hmac.compare_digest, which rejects non-ASCII str operands. On an auth endpoint this can turn a forged request into an unhandled 500 instead of a clean rejection (False / 401).

This is a robustness issue on the error path, not an authentication bypass: no invalid hash is accepted.

### Steps to reproduce

from aiogram.utils.web_app import check_webapp_signature

TOKEN = "123456:ABC-DEF1234ghIkl-zyx57W2v1u123ew11"
check_webapp_signature(TOKEN, "id=1&hash=%C3%A9%C3%A9")
# TypeError: comparing strings with non-ASCII characters is not supported

### Code example

```python3
from aiogram.utils.web_app import check_webapp_signature

TOKEN = "123456:ABC-DEF1234ghIkl-zyx57W2v1u123ew11"
check_webapp_signature(TOKEN, "id=1&hash=%C3%A9%C3%A9")
```

### Logs

```sh
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "aiogram/utils/web_app.py", line 140, in check_webapp_signature
    return hmac.compare_digest(calculated_hash, hash_)
TypeError: comparing strings with non-ASCII characters is not supported
```

### Additional information

Suggested fix:

return hmac.compare_digest(calculated_hash.encode(), hash_.encode())

Happy to open a PR with tests.

nvikou:
## Summary

- Fixes #1903
- `check_webapp_signature` raised `TypeError` on non-ASCII client-supplied `hash` instead of returning `False`
- Compare digests as bytes so `hmac.compare_digest` accepts any input and keeps the `-> bool` contract

## Test plan

- [x] New tests in `tests/test_utils/test_web_app.py` for non-ASCII / percent-encoded / empty hash
- [ ] Existing web app signature tests still pass on CI

### DIFF
--- a/CHANGES/1903.bugfix.rst
+++ b/CHANGES/1903.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed :code:`check_webapp_signature` raising ``TypeError`` instead of returning ``False`` when the client-supplied hash contains non-ASCII characters.
--- a/aiogram/utils/web_app.py
+++ b/aiogram/utils/web_app.py
@@ -137,7 +137,7 @@ def check_webapp_signature(token: str, init_data: str) -> bool:
         msg=data_check_string.encode(),
         digestmod=hashlib.sha256,
     ).hexdigest()
-    return hmac.compare_digest(calculated_hash, hash_)
+    return hmac.compare_digest(calculated_hash.encode(), hash_.encode())
 
 
 def parse_webapp_init_data(
--- a/tests/test_utils/test_web_app.py
+++ b/tests/test_utils/test_web_app.py
@@ -86,3 +86,36 @@ class TestWebApp:
                 "&query_id=test"
                 "&hash=test",
             )
+
+
+class TestCheckWebappSignatureInvalidHash:
+    @pytest.mark.parametrize(
+        "hash_value",
+        [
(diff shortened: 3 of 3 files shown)


nvikou:
The changelog fragment is already in the PR (`CHANGES/1903.bugfix.rst`).
The "Check that changes are described" job fails at checkout with:
"Refusing to check out fork pull request code from a pull_request_target workflow"
— so towncrier never runs. Happy to adjust if needed.

Olegt0rr (approved):

PR #1874, opened 2026-08-10
docs: add Python version requirement to install.rst

Strictman:
# Description

`docs/install.rst` did not mention the minimum required Python version, while the README states that aiogram requires Python 3.10+. This adds a short "Requirements" section at the top of the installation guide, so new users see this requirement before installing.

Fixes # (no linked issue, small documentation improvement)

## Type of change

Please delete options that are not relevant.
- [x] Documentation (typos, code examples or any documentation update)

# How Has This Been Tested?

Verified the `.rst` file renders correctly by checking that heading underline lengths match the headings and that the syntax is valid reStructuredText.

**Test Configuration**:
* Operating System: macOS 26.5.1 (25F80)
* Python version: N/A (documentation-only change, not tested against a live install)

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #1908, opened 2026-09-08
CI: harden and modernise the GitHub Actions workflows (zizmor audit, Node 24, lint/build/docs jobs)

### LINKED ISSUE #1906: CI: GitHub removes Node 20 from Actions runners on 2026-09-23; several of our actions still declare it
### What GitHub announced

From the GitHub changelog ["Deprecation of Node 20 on GitHub Actions runners"](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/):

- "Beginning on June 16th, 2026, runners will begin using Node24 by default."
- Actions that declare `runs.using: node20` are then forced onto Node 24; the only way to keep Node 20 is the environment variable `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true`.
- "This will only work until we upgrade the runner and remove Node20 on September 23rd, 2026."

So since June 16 our workflows already execute Node 20 actions on Node 24, and after **September 23, 2026** the escape hatch disappears: if any of those action versions misbehaves on Node 24, there is no way to pin back except upgrading the action.

### What is affected on `dev-3.x` today

Actions in `.github/workflows` whose pinned major still declares Node 20 (checked in each action's `action.yml`):

| Workflow | Action | Pinned | Node 24 version |
|---|---|---|---|
| `tests.yml` (CPython job), `pull_request_changelog.yml` | `actions/checkout` | `v4` | `v7` |
| `tests.yml` (CPython job), `pull_request_changelog.yml` | `actions/setup-python` | `v5` | `v7` (or drop it: `setup-uv` installs Python) |
| `pull_request_changelog.yml` | `peter-evans/find-comment` | `v3` | `v4` |
| `pull_request_changelog.yml` | `peter-evans/create-or-update-comment` | `v4` | `v5` |
| `label_pr.yaml` | `andymckay/labeler` | `master` | none: repository archived in 2022, declares Node 12 |

Everything else (`checkout@v6`, `setup-python@v6`, `setup-uv@v7`, `upload-artifact@v6`, `download-artifact@v7`, `codecov-action@v5` composite, `pypa/gh-action-pypi-publish`) already runs on Node 24 or is not a JavaScript action.

### LINKED ISSUE #1907: CI: zizmor reports 40 findings on the GitHub Actions workflows of `dev-3.x`
[zizmor](https://docs.zizmor.sh/) is a static analyser for GitHub Actions workflows. Run against the four workflows in `.github/workflows` on `dev-3.x` (zizmor 1.30.0):

```bash
uvx zizmor .github/workflows
uvx zizmor --persona=pedantic .github/workflows
```

Regular persona: 40 findings shown (55 total, 15 suppressed as low-confidence): 27 high, 11 medium, 2 low. Pedantic persona adds 15 more, all informational or low.

### High

**`label_pr.yaml`: `pull_request_target` runs a mutable third-party ref with a write token** (`dangerous-triggers`, `unpinned-uses`, `excessive-permissions`). The job has no `permissions:` block, so it gets the default write-capable `GITHUB_TOKEN`, and it runs `andymckay/labeler@master`: a moving branch of a repository that was archived in September 2022 and whose `action.yml` declares Node 12. A compromise of that repository or account would hand a write token for `aiogram/aiogram` to whoever controls `master`.

**`pull_request_changelog.yml`: `pull_request_target` plus checkout of the PR head** (`dangerous-triggers`, `artipacked`, `excessive-permissions`). The workflow checks out `github.event.pull_request.head.sha` under the privileged trigger with default permissions and `persist-credentials` left at its default `true`, so the write token is written into `.git` next to contributed code. Today the only thing executed on that checkout is `towncrier check`, which reads `pyproject.toml` and does not import the package or render templates, so there is no direct code execution from PR content; any future step that installs or runs the project would turn this into a token leak. This is the same trigger design that makes the check fail on every fork PR today (see #1884: `actions/checkout` now refuses fork code under `pull_request_target`).

**`pypi-release.yml:25`: template injection into `run:`** (`template-injection`). `${{ github.ref }}` is expanded directly into the shell command that derives the version. The workflow only triggers on `v3.*` tags, so the practical exposure is low, but the pattern is exactly what the audit exists for.

**`pypi-release.yml:21`: shared cache in the release build** (`cache-poisoning`, low confidence). `astral-sh/setup-uv` caches by default, and the job that builds the artifacts published to PyPI restores that cache.

**23 × `unpinned-uses`**: every `uses:` in the four workflows is pinned to a tag, not a commit SHA. zizmor's default policy treats that as high severity. Of these, only `andymckay/labeler@master` points at a mutable branch; the rest are version tags of GitHub, Astral, Codecov, PyPA and peter-evans actions.

### Medium

**No `permissions:` block anywhere** (`excessive-permissions`, 9 findings across all four workflows): every job runs with the repository default token scope, including the release job that only needs `id-token: write` on its publish step and the test matrix that needs read access only.

**`actions/checkout` without `persist-credentials: false`** (`artipacked`, 4 findings: `tests.yml:54` and `:127`, `pypi-release.yml:13`, `pull_request_changelog.yml:16`). None of these jobs uses the token after checkout.

### Low and pedantic only

- No `concurrency:` in any workflow (`concurrency-limits`, 4): superseded PR runs keep running to completion.
- Jobs `tests`, `pypy-tests`, `put-label` and the changelog jobs have no `name:` (`anonymous-definition`, 5).
- `id-token: write` in `pypi-release.yml` has no explanatory comment (`undocumented-permissions`).
- `peter-evans/create-or-update-comment` could be replaced by `gh pr comment` (`superfluous-actions`, 3).

### Full finding list (regular persona)

| Workflow | Audit | Location |
|---|---|---|
| `label_pr.yaml` | `dangerous-triggers` | line 3 |
| `label_pr.yaml` | `excessive-permissions` | line 11 |
| `label_pr.yaml` | `unpinned-uses` | line 15 |
| `pull_request_changelog.yml` | `dangerous-triggers` | line 2 |
| `pull_request_changelog.yml` | `artipacked` | line 16 |
| `pull_request_changelog.yml` | `excessive-permissions` | lines 1, 12, 78 |
| `pull_request_changelog.yml` | `unpinned-uses` | lines 17, 23, 39, 48, 68, 83, 91 |
| `pypi-release.yml` | `template-injection` | line 25 |
| `pypi-release.yml` | `cache-poisoning` | line 21 |
| `pypi-release.yml` | `artipacked` | line 13 |
| `pypi-release.yml` | `excessive-permissions` | lines 1, 9 |
| `pypi-release.yml` | `unpinned-uses` | lines 13, 16, 21, 39, 55, 61 |
| `tests.yml` | `artipacked` | lines 54, 127 |
| `tests.yml` | `excessive-permissions` | lines 1, 23, 108 |
| `tests.yml` | `unpinned-uses` | lines 55, 58, 63, 79, 85, 100, 128, 131, 136 |

Olegt0rr:
Closes #1906, closes #1907. Supersedes #1884: its three commits are included as-is at the base of this branch (with their changelog fragment), and the changelog workflows are hardened on top of them.

## What this does

A full pass over `.github/workflows` driven by a [zizmor](https://docs.zizmor.sh/) audit (#1907) and GitHub's removal of Node 20 from runners on 2026-09-23 (#1906). One commit per workflow so each file can be reviewed as a whole.

**`tests.yml`**: lint moves out of the 15-leg matrix into its own `Lint` job that runs `make lint` (ruff, mypy on Linux and with `--platform win32`, pre-commit hooks); PyPy 3.11 joins the main matrix; new `Build package` job (`uv build`, `twine check --strict`, wheel and sdist smoke-installed with `python -P`); new `Docs` job (Sphinx build, `viewcode` skipped in CI only, non-strict until the existing warnings are fixed). Redis and MongoDB run on Linux legs only: Redis from `tests/docker-compose.yml`, MongoDB from MongoDB's apt repo. `actions/setup-python` is gone, `setup-uv` installs the interpreter; every `uv sync` is `--locked`; path filters are removed (they missed `uv.lock`). `Build package` and `Docs` start alongside `Lint` and `Tests`, nothing is gated.

**`pypi-release.yml`**: `permissions: {}` with `id-token: write` only on publish; the `github.ref` template injection is fixed; the build asserts the installed package version equals the tag; the environment URL now works (it referenced a step of another job); no shared cache on the release build.

**`pull_request_changelog.yml` + `pull_request_changelog_comment.yml`**: the #1884 split (check on `pull_request`, comment from `workflow_run`), plus: fetch separated from the towncrier step so infra failures read as `inconclusive` rather than "changelog missing", result mirrored into the job summary, artifact `overwrite` for re-runs, commenter only acts on `success`/`failure` runs and tolerates a missing artifact.

**`label_pr.yaml`**: archived `andymckay/labeler@master` (Node 12) replaced by `actions/labeler@v7` with `.github/labeler.yml`.

**New**: `workflow-lint.yml` (zizmor with `.github/zizmor.yml`, actionlint, and `scripts/check_action_refs.py`, a YAML-aware check that rejects `uses:` refs that are not version tags or commit SHAs), `dependabot.yml` for actions (weekly, grouped, 7-day cooldown, labelled `skip news`).

**Alignment**: `make lint` and `make lint-workflows` are what CI runs; AGENTS.md, CLAUDE.md and the checked-in agent instructions describe the new CI; `codecov.yaml` waits for all 15 uploads; the dead `--cov-config .coveragerc` flag is gone.

Deliberate non-goals: no SHA pinning (tags plus the ref guard), Redis/Mongo not on macOS/Windows, docs job without `-W` yet.

## Decisions

The trade-offs behind the changes, kept from the working plan so reviewers do not have to reconstruct them.

| Decision | Choice | Why |
|---|---|---|
| Action pinning | Version tags (`@v7`, `@v10.0.1`), no commit SHAs | SHA pins turn every upstream patch release into a Dependabot PR that someone has to review and merge, and the security they add only materialises if the reviewer actually inspects the upstream commit behind each new hash, which nobody does at this project's maintainer bandwidth. Version tags from GitHub, Astral, Codecov and PyPA are trusted enough for this repository; the guard is `zizmor` with the `ref-pin` policy, which still rejects branch refs like `@master`. |
| `label_pr.yaml` | Rewrite with the official `actions/labeler@v7`, keep `pull_request_target` | `andymckay/labeler` is dead: the repository was archived in September 2022, its `action.yml` declares Node 12, and it is referenced by the mutable `@master` ref while holding a write token. The `3.x` label is what makes `label:3.x` return issues and PRs together (518 of 561 PRs to `dev-3.x` carry it). Fork PRs get a read-only token under `pull_request`, so labelling needs `pull_request_target`; safe because the job never checks out or runs PR content. |
| Lint | Separate `lint` job, runs once | Today ruff + mypy + format run inside all 15 matrix jobs. Measured: the whole lint is ~6 s (mypy 2.0 with `--native-parser` is 3 s), so this is not about runner minutes; it is about one clearly named red check instead of 15, and about running the pre-commit hooks exactly once. |
| PyPy | Folded into the main matrix via `exclude` | The two job definitions share 34 of 40 lines. Coverage and Codecov are skipped for PyPy. |
| Redis / MongoDB | Linux only. Redis via `docker compose -f tests/docker-compose.yml up -d redis` (`redis:8-alpine`), MongoDB via `mongodb-org-server` deb from MongoDB's apt repo (43 MB vs 300 MB for the Docker image) | The servers are black boxes on the other end of a socket; OS-specific code under test is aiogram + its Python deps, which macOS / Windows still exercise. `mongod --fork` returns only once it listens, so no wait loop. Wall time is a wash: the Docker pull took 12 s on the last run, `apt-get update` + install is in the same range; the win is bandwidth and no Docker dependency. Without `mongosh` there is nobody to create a user, so the URI drops credentials; `conftest.py` only forwards the string. |
| `actions/setup-python` | Removed everywhere | `setup-uv` installs the interpreter itself via `python-version`, CPython and PyPy (`pypy3.11` is a valid uv request). |
| Changelog check and its PR comment | Both kept, on the split design from #1884: the check runs on `pull_request` (read-only token) and uploads a one-word state artifact; a separate `workflow_run` workflow with `pull-requests: write` posts the comment without ever checking out PR code. The check additionally writes the same text to the job summary. | The check is mandatory and currently broken: under `pull_request_target`, `actions/checkout` refuses fork PR code, so `towncrier check` never runs and the "Changelog is required" comment is posted spuriously on every fork PR (#1874, #1878, #1879, #1881, #1883 and #1884 itself). #1884 splits the workflow along the privilege boundary, which removes the `pull_request_target` + checkout finding and keeps the contributor-facing comment. `workflow_run` is also flagged by zizmor as a dangerous trigger; it is acceptable here because the privileged half validates the artifact against an allowlist, resolves the PR number from the API rather than from the artifact, and runs no contributed code. |
| Path filters on `tests.yml` | Removed | They already missed `uv.lock`, `examples/**` and `scripts/**`, and path-filtered required checks block PRs that do not match the filter. The whole suite is minutes; run it always. If the maintainers want filtering back, `paths-ignore` on `**.md`, `**.rst`, `CHANGES/**` is the safe subset. |
| Docs in CI | New `docs` job, non-strict build, `-j auto`, starts immediately (no `needs`) | Nothing validates the Sphinx build before merge today. Measured: 685 rst files, 299 s serial and 162 s with `-j auto` locally, so 3–5 min on a 4-vCPU runner. That makes docs the longest job by far (the whole test matrix is ~2 min), so it must start at t=0; putting it behind `lint` would add lint's ~40 s of setup to the PR wall time. The build emits 71 warnings (6 docutils errors), so `-W` cannot be enabled yet. Follow-up issue: fix the warnings, then add `-W`. |
| `setup-uv` version | Pin `@v10.0.1` exactly | Astral stopped moving major tags after v7; `@v10` does not exist. Dependabot keeps it current. |
| `uv sync` | Always `--locked`, always `--no-extra docs` outside the docs job | `--locked` fails when `uv.lock` drifts from `pyproject.toml` instead of silently re-locking. The `docs` extra is Sphinx + theme + i18n tooling, useless to tests and lint. |
| `--cov-config .coveragerc` | Removed (also from `Makefile`) | The file does not exist; coverage config lives in `pyproject.toml`. |
| `codecov.yaml` `after_n_builds` | 6 → 15 | There are 15 coverage uploads (5 CPython × 3 OS); the PR comment currently posts after 6. |
| `concurrency` on `tests.yml` | `cancel-in-progress` only for PRs | Cancelling `dev-3.x` runs on back-to-back merges leaves Codecov without a base report. |

## Validation

Locally: `uvx zizmor --config .github/zizmor.yml .` clean (pedantic: only three informational hints about `peter-evans` steps), `actionlint` clean, `pre-commit run --all-files` green, `make lint` and `make lint-workflows` exit 0, full test suite 2313 passed / 51 skipped in an environment synced exactly like the Tests job, docs build exit 0 (35 s without viewcode vs 145 s with), release smoke loop verified against a real `uv build` output including a deliberate version mismatch.

Cannot be verified before merge: `pull_request_changelog_comment.yml` and `label_pr.yaml` run from the base branch (`workflow_run` / `pull_request_target`), so the first PR after merge exercises them; `pypi-release.yml` runs on the next `v3.*` tag. Branch protection required-check names change to `Lint`, `Tests (<os>, <python>)`, `Build package`, `Docs`, `Changelog fragment` and need updating if any of the old names are required.

Follow-ups: #1909 (the Sphinx build emits 71 warnings, 59 of them from the towncrier template via `CHANGES.rst`; once fixed the `Docs` job gets `-W`); `ruff check` does not cover `tests/` in CI while the pre-commit hook does (12 preview-rule errors there today, not filed yet).


scripts/check-action-refs.sh
```diff
@@ -0,0 +1,18 @@
...
+# silent: it checks that a ref is present, not what the ref points at.
+set -euo pipefail
+
+cd "$(dirname "$0")/.."
+
+# `pypa/gh-action-pypi-publish@release/v1` is the one documented exception (the ref PyPA
+# recommends). Composite actions under `.github/actions/**` are covered too, hence the recursion.
+if grep -rnE --include='*.yml' --include='*.yaml' '^\s*(-\s+)?uses:\s*\S+@' .github \
+     | grep -vE '@(v?[0-9]+(\.[0-9]+)*|[0-9a-f]{40})([[:space:]]|$)' \
+     | grep -vE 'pypa/gh-action-pypi-publish@release/v[0-9]+([[:space:]]|$)'; then
```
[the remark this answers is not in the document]
Olegt0rr: Switched the pattern to POSIX classes (`[[:space:]]` / `[^[:space:]]`) in 24cca7ca. For the record, GNU and BSD grep both accept `\s`/`\S` in ERE, so the check was not missing lines, but the POSIX form is the portable one. Re-tested: silent on the tree, flags `-   uses: foo/bar@main` in a composite action.

.github/workflows/workflow-lint.yml
```diff
@@ -0,0 +1,49 @@
+name: Workflow lint
+
+on:
+  push:
+    branches:
+      - dev-3.x
+    paths:
+      - ".github/**"
+      - ".pre-commit-config.yaml"
```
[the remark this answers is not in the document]
Olegt0rr: Added `scripts/check-action-refs.sh` to both `paths:` lists (push and pull_request) in 24cca7ca, so edits to the guard run Workflow lint themselves.

scripts/check-action-refs.sh
```diff
@@ -0,0 +1,18 @@
...
+# reject refs that do not even look like a version tag (`v?1.2.3`) or a 40-hex SHA - `main`,
+# `master`, `release/x`, `2.x`, `v4-beta`. That is exactly where zizmor's `ref-pin` policy is
+# silent: it checks that a ref is present, not what the ref points at.
+set -euo pipefail
+
+cd "$(dirname "$0")/.."
+
+# `pypa/gh-action-pypi-publish@release/v1` is the one documented exception (the ref PyPA
+# recommends). Composite actions under `.github/actions/**` are covered too, hence the recursion.
+if grep -rnE --include='*.yml' --include='*.yaml' '^[[:space:]]*(-[[:space:]]+)?uses:[[:space:]]*[^[:space:]]+@' .github \
```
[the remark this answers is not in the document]
Olegt0rr: Fixed in 7cec499e: the pattern is now `uses[[:space:]]*:[[:space:]]*...`, so `uses : owner/action@main` is caught. Tested with a probe workflow (flagged) and the real tree (silent). The same commit also tightens the commenter's PR lookup per the suppressed note: it now requires `state === "open"` plus the run's `head_branch` and `head_repository.full_name`, not just the SHA.

.github/workflows/pull_request_changelog.yml
```diff
@@ -1,99 +1,111 @@
...
     if: "!contains(github.event.pull_request.labels.*.name, 'skip news')"
     steps:
-      - name: Checkout code
-        uses: actions/checkout@v4
+      - uses: actions/checkout@v7
         with:
+          # the default merge ref does not exist for a PR with merge conflicts, so the checkout
+          # would fail; the head commit always exists, and under `pull_request` the token is
+          # read-only, so checking out contributed code here is safe
           ref: ${{ github.event.pull_request.head.sha }}
```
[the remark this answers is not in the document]
Olegt0rr: You are right: GitHub documents that "Workflows will not run on pull_request activity if the pull request has a merge conflict", so the head.sha checkout never delivered that guarantee. Reverted to the default merge-ref checkout and dropped the comment and the PR-description claim in d2844284. The inconclusive-state handling stays for the other failure modes (runner loss, fetch errors).

scripts/check-action-refs.sh
```diff
@@ -0,0 +1,18 @@
...
+# silent: it checks that a ref is present, not what the ref points at.
+set -euo pipefail
+
+cd "$(dirname "$0")/.."
+
+# `pypa/gh-action-pypi-publish@release/v1` is the one documented exception (the ref PyPA
+# recommends). Composite actions under `.github/actions/**` are covered too, hence the recursion.
+if grep -rnE --include='*.yml' --include='*.yaml' '^[[:space:]]*(-[[:space:]]+)?uses[[:space:]]*:[[:space:]]*[^[:space:]]+@' .github \
+     | grep -vE '@(v?[0-9]+(\.[0-9]+)*|[0-9a-f]{40})([[:space:]]|$)' \
+     | grep -vE 'pypa/gh-action-pypi-publish@release/v[0-9]+([[:space:]]|$)'; then
```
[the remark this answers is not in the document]
Olegt0rr: Replaced the grep script with a YAML-aware `scripts/check_action_refs.py` in d2844284 (PyYAML via inline script metadata, run with `uv run`, shared by CI and `make lint-workflows`). It walks every `uses` key in `.github/**/*.yml|yaml` after parsing, so quoted keys, quoted values, flow mappings, composite actions and reusable-workflow jobs are all handled. Fixture with 14 spellings: `"uses": x@main`, `uses : x@main`, `{uses: x@master}`, `@2.x`, `@v4-beta`, non-pypa `@release/v1` and a composite action are rejected; `uses: "x@v1"`, a reusable workflow `@v3`, a 40-hex SHA, `docker://`, `./local` and the pypa exception pass.

scripts/check_action_refs.py
```diff
@@ -0,0 +1,77 @@
...
+            else:
+                yield from iter_uses(value)
+    elif isinstance(node, list):
+        for item in node:
+            yield from iter_uses(item)
+
+
+def is_pinned(uses: str) -> bool:
+    if "@" not in uses or uses.startswith(("./", "docker://")):
+        return True  # local action or container image: nothing to pin
```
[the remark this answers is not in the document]
Olegt0rr: Fixed in 0e663238: only `./` local actions and `docker://` images bypass the ref check now; a remote `owner/action` without `@` is reported as unpinned. Added `foo/noref` to the fixture (rejected). The same commit also pins twine exactly (`twine==7.0.0`) in both workflows per the suppressed notes, so the "bump deliberately" comment is true.

.github/zizmor.yml
```diff
@@ -0,0 +1,8 @@
+rules:
+  unpinned-uses:
+    config:
+      policies:
+        "*": ref-pin   # accepts any tag or branch; the workflow-lint grep step is what rejects refs that are not version tags or commit SHAs (SHA pins would mean reviewing every Dependabot bump by hand, see plan, decisions table)
```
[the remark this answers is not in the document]
Olegt0rr: Updated in 0e663238: the comment now points at `scripts/check_action_refs.py` (Workflow lint, `make lint-workflows`) instead of the removed grep step.

.github/workflows/pull_request_changelog_comment.yml
```diff
@@ -0,0 +1,137 @@
...
+        if: steps.result.outputs.state
+        id: pr
+        uses: actions/github-script@v8
+        with:
+          script: |
+            const run = context.payload.workflow_run;
+            const headSha = run.head_sha;
+            const headBranch = run.head_branch;
+            const headRepo = run.head_repository?.full_name;  // null if the fork was deleted
+            const { data: prs } = await github.rest.repos.listPullRequestsAssociatedWithCommit({
```
JrooTJunior, line 67: **Blocker:** `listPullRequestsAssociatedWithCommit` returns nothing for fork commits when you query the base repo. Tried it on this PR's head:

```
GET repos/aiogram/aiogram/commits/0e663238…/pulls   -> []
GET repos/Olegt0rr/aiogram/commits/0e663238…/pulls  -> [#1908]
```

So every fork PR logs "No pull request found" and never gets a comment, and fork PRs are exactly what #1884 set out to fix. Looking the PR up by head works for forks:

```js
const { data: prs } = await github.rest.pulls.list({
  owner: context.repo.owner,
  repo: context.repo.repo,
  state: "open",
  head: `${run.head_repository.owner.login}:${run.head_branch}`,
});
const pr = prs.find((item) => item.head.sha === headSha);
```

(`head_repository` can be null if the fork was deleted, so keep the guard.)

.github/workflows/pull_request_changelog.yml
```diff
@@ -1,99 +1,107 @@
-name: "Check that changes are described"
+name: Check that changes are described
+
 on:
-  pull_request_target:
+  pull_request:
```
JrooTJunior, line 4: Just noting this, no change needed: under `pull_request`, the workflow file comes from the PR merge ref, so a fork can edit this file and make it upload `success`. The damage is limited to a wrong comment and a green check on that PR, and it's the price of moving off `pull_request_target`. It does mean a PR that touches `.github/**` needs a careful look before the check can be trusted.

.github/workflows/tests.yml
```diff
@@ -34,114 +69,147 @@ jobs:
...
-      matrix:
-        os:
-          - ubuntu-latest
-          - macos-latest
-          # - windows-latest
-        python-version:
-          - "pypy3.11"
+  build:
+    name: Build package
+    needs: lint
```
JrooTJunior, line 156: `build` doesn't use anything from `lint`, so `needs: lint` only delays it by lint's setup time. That's the reason you gave for not gating `docs`. Drop it?

Makefile
```diff
@@ -37,9 +37,19 @@ install: clean
...
 .PHONY: lint
 lint:
-	uv run ruff format --check --diff $(package_dir)
+	uv run ruff format --check --diff $(package_dir) $(tests_dir) $(scripts_dir) $(examples_dir)
 	uv run ruff check --show-fixes --preview $(package_dir) $(examples_dir)
 	uv run mypy --native-parser --num-workers 8 $(package_dir)
+	# aiogram/types/custom.py has a `sys.platform == "win32"` branch that mypy prunes as
+	# unreachable on Linux; this second pass type-checks it.
+	uv run mypy --native-parser --num-workers 8 --platform win32 $(package_dir)
+	uv run pre-commit run --all-files --show-diff-on-failure
```
JrooTJunior, line 46: `pre-commit run --all-files` runs fixers (`trailing-whitespace`, `end-of-file-fixer`, `ruff-format`), so `make lint` now edits files locally. AGENTS.md calls it the "Quick loop" check, which suggests it's read-only. Either mention in AGENTS.md that it can modify files, or leave pre-commit out of the target and call it separately in CI.

JrooTJunior (changes requested):
Thanks for this. Locally, for `0e663238`: `check_action_refs.py`, `actionlint`, `zizmor` (online), `uv lock --check` and CI-style `make lint` all pass, and `uv build` + `twine check --strict` + the wheel/sdist smoke install work.

One blocker is inline: the commenter can't resolve fork PRs.

Before merge:
- The new `Tests` / `Workflow lint` runs are waiting on approval because this is a fork PR. I'll approve them so the new matrix gets a real run: PyPy on Linux with Redis/Mongo, Mongo from apt without auth, and the Docs job.
- Required check names change (`Lint`, `Tests (<os>, <python>)`, `Build package`, `Docs`, `Changelog fragment`), so branch protection has to be updated when this merges.

Small things:
- The PR body says Redis runs via `docker run redis:8-alpine`, but the workflow uses `docker compose`.
- `enable-cache: true` does nothing, since `setup-uv` already caches by default on GitHub-hosted runners.

.github/workflows/pull_request_changelog_comment.yml
Olegt0rr: Thanks, confirmed and fixed in 8afd1fd4: the commenter now uses `pulls.list({state: "open", head: `${run.head_repository.owner.login}:${run.head_branch}`})` and picks the PR whose current head is the run's `head_sha`; the null guard for a deleted fork is kept and skips with a log line. The old `listPullRequestsAssociatedWithCommit` lookup is gone.

.github/workflows/pull_request_changelog.yml
Olegt0rr: Agreed, no change made. That is the trade-off of `pull_request` over `pull_request_target`, and the blast radius is one wrong comment plus a green check on that PR only; the artifact is still allowlisted and the PR number comes from the API. A PR that edits `.github/**` needs a human look before its changelog check is trusted.

.github/workflows/tests.yml
Olegt0rr: Dropped `needs: lint` on `build` in 8afd1fd4; it now starts with `lint`, `tests` and `docs`. The PR description no longer claims the gating.

Makefile
Olegt0rr: Went with the documentation option in 8afd1fd4: AGENTS.md now says `make lint` runs the pre-commit fixers (`trailing-whitespace`, `end-of-file-fixer`, `ruff-format`) and may modify the working tree, and asks contributors to review `git diff` afterwards. Kept pre-commit inside the target so local and CI stay one command; happy to split it out instead if you prefer a read-only `make lint`.

Olegt0rr:
Addressed the review in 8afd1fd4.

**Changes**
- Commenter looks the PR up with `pulls.list(head="owner:branch")`, which works for fork PRs; the old `listPullRequestsAssociatedWithCommit` lookup returned nothing for fork commits (the blocker).
- `needs: lint` dropped on `build`; it starts alongside `lint`, `tests` and `docs`.
- The no-op `enable-cache: true` lines removed from the `setup-uv` steps.
- AGENTS.md notes that `make lint` runs the pre-commit fixers and may modify files.
- PR description corrected: Redis comes from `docker compose -f tests/docker-compose.yml`, not `docker run`.

**Status**
- The red `changes-required` check is the base-branch `pull_request_target` workflow refusing the fork checkout, the #1884 bug this PR fixes; it keeps failing until merge.
- Ready for the workflow-run approval whenever convenient.

PR #1911, opened 2026-09-12
Fix ephemeral reply addressed to the bot instead of receiver_user

### LINKED ISSUE #1897: `Message.as_ephemeral_message_parameters()` addresses the reply to the bot itself instead of `receiver_user`
## Checklist

- [x] I am sure the error is coming from aiogram code and not from my code

## Context

Bot API 10.2/10.3 define two shapes of ephemeral message:

1. **Ephemeral command (user → bot)**: incoming `Message` where `from_user` is the human.
2. **Ephemeral message (bot → user)**: the `Message` returned by `send_*` with `ephemeral_message_parameters=...`, where `from_user` is the **bot** and the human is `Message.receiver_user` ("For ephemeral messages, the user who received the message").

The new helper introduced in 3.31.0 ([aiogram/types/message.py#L910-L925](https://github.com/aiogram/aiogram/blob/v3.31.0/aiogram/types/message.py#L910-L925)) always uses `from_user`:

```python
return EphemeralMessageParameters(
    receiver_user_id=self.from_user.id,
    ...
)
```

For shape 2 this is the bot's own id. The generated `edit_ephemeral_*` / `delete_ephemeral` shortcuts on the very same object use `receiver_user_id=self.receiver_user.id`, so two shortcuts on one `Message` resolve the same API field to two different users.

The pre-3.31 inline expression (`self.from_user.id if self.ephemeral_message_id and self.from_user else None`) had the same defect; 3.31.0 turned it into a public documented helper and pinned it with tests (`test_as_ephemeral_message_parameters`, `test_reply_to_ephemeral_message_is_ephemeral` assert `== message.from_user.id`).

## Reproduction

```python
import datetime
from aiogram.types import Chat, Message, User

# Shape of the Message returned by bot.send_message(..., ephemeral_message_parameters=...)
# This is also the repo's own TEST_MESSAGE_EPHEMERAL fixture.
m = Message(
    message_id=0,
    date=datetime.datetime.now(),
    text="hi",
    chat=Chat(id=-100, type="supergroup"),
    from_user=User(id=42, is_bot=True, first_name="Bot"),
    receiver_user=User(id=43, is_bot=False, first_name="Alice"),
    ephemeral_message_id=7,
)

print(m.reply("pong").ephemeral_message_parameters.receiver_user_id)  # 42  <- the bot
print(m.edit_ephemeral_text("x").receiver_user_id)                     # 43  <- Alice
```

## Expected behavior

`m.reply(...)` on a bot-sent ephemeral message addresses the reply to `receiver_user` (43), exactly like `edit_ephemeral_text` does.

## Current behavior

The reply is addressed to the bot's own user id; Telegram either rejects it or the human never sees it.

## Suggested fix

```python
receiver_user_id=(self.receiver_user or self.from_user).id
```

This keeps shape 1 (incoming ephemeral command, `receiver_user` absent) working and fixes shape 2. Update the two tests that pin `from_user.id`.

## aiogram version

3.31.0

TanbirRamim:
# Description

`Message.as_ephemeral_message_parameters()` always used `from_user`, so calling `reply()` on an ephemeral message the bot sent addressed the reply to the bot itself, while `edit_ephemeral_*` on the same message used `receiver_user`. It now uses `receiver_user` when it is a non-bot user and falls back to `from_user` for incoming ephemeral commands (ephemeral messages are never shown to bots, so a bot `receiver_user` is skipped either way).

Fixes #1897

## Type of change

# How Has This Been Tested?

Ran the reproduction from the issue (now prints 43 for both calls), updated the two tests that pinned `from_user.id` and added cases for incoming ephemeral commands. `pytest tests/test_api/test_types` passes (484), along with `ruff check`, `ruff format --check`, `mypy aiogram` and `towncrier check`.

**Test Configuration**:
* Operating System: macOS
* Python version: 3.12


TanbirRamim:
The changelog fragment is included (`CHANGES/1897.bugfix.rst`, named after issue #1897) and `towncrier check --compare-with origin/dev-3.x` passes locally. The `changes-required` job failed at its "Checkout code" step before towncrier ran, so it probably just needs a re-run.

PR #1914, opened 2026-09-17
Fix check_webapp_signature raising TypeError on non-ASCII hash

### LINKED ISSUE #1903: check_webapp_signature raises TypeError on non-ASCII hash instead of returning False

cristianchiriac:
Fixes #1903.

## Bug

```python
from aiogram.utils.web_app import check_webapp_signature

TOKEN = "123456:ABC-DEF1234ghIkl-zyx57W2v1u123ew11"
check_webapp_signature(TOKEN, "id=1&hash=%C3%A9%C3%A9")
# TypeError: comparing strings with non-ASCII characters is not supported
```

`check_webapp_signature` is annotated `-> bool` and every other invalid-input path returns `False`. But when the client-supplied `hash` parameter contains non-ASCII characters (e.g. after percent-decoding `%C3%A9` to `é`), `hmac.compare_digest()` raises `TypeError` instead — it rejects non-ASCII `str` operands. On an auth endpoint this turns a forged/malformed request into an unhandled 500 instead of a clean rejection.

This is a robustness issue on the error path, not an authentication bypass: no invalid hash was ever accepted.

## Fix

`calculated_hash` is always a hex digest (ASCII-safe); `hash_` is the untrusted one. Compare both as UTF-8-encoded bytes instead of `str` — `hmac.compare_digest` has no ASCII restriction on bytes operands, and any `str` can always be UTF-8 encoded, so this can't itself raise.

## Test plan

- Added the reported non-ASCII case to the existing `test_check_webapp_signature` parametrization, asserting it returns `False`.
- Verified the test fails with the exact reported `TypeError` against the unpatched code, and passes after the fix.
- `uv run ruff check --preview` / `uv run ruff format --check` / `uv run mypy aiogram` — all clean.
- `uv run pytest tests` — 2310 passed, 55 skipped (same skip count as before this change).
- Added `CHANGES/1903.bugfix.rst`, verified with `towncrier build --draft`.


Olegt0rr:
@Tatamis, PR already exists: #1904

Tatamis:
Ah, missed that — thanks for pointing it out. Closing this one in favor of #1904.

